PR #5, opened 2024-05-06
Refactor code that pulls reference tree and wrap code in a CLI

### LINKED ISSUE #1: Support for historical versions of reference trees (along with their corresponding root sequence)
We should be able to pull and operate with past versions of reference trees.  Note that much of the content of this issue was provided by John Huddleston.

My understanding is that we can pull the latest version of a reference tree that was available as of a given date using a command like the following:

```
curl https://nextstrain.org/nextclade/sars-cov-2@2023-10-26 \
  --header 'Accept: application/vnd.nextstrain.dataset.main+json' \
  --compressed > tree_2023-10-26.json
```

The key here is the `@2023-10-26` in the first line, which indicates that we want the latest tree that was available as of that date.

When we do this, we may need to also pull the root reference sequence. There are two cases here. For newer trees, that reference sequence will be stored directly in the tree's json file at `json["root_sequence"]["nuc"]`. That can be pulled out into a `.fasta` file as follows:

```
jq -r '.root_sequence .nuc' tree.json > root_sequence.fasta
```

For older trees, that reference root sequence will not be stored directly in the tree's json.  I think for our purposes, it's probably ok to ignore that -- we're doing this process going forward and don't need to look back to older trees.

bsweger:
Resolves #1

This PR mixes a few concerns (sorry!), so reviewing commit-by-commit probably isn't super helpful.

1. Adds a third-party logger (I'm so over Python's native logger)
2. Creates a separate function that retrieves a reference tree for a specific date (assumes the response will also contain root_sequence_information)
3. Wraps everything in a little CLI so it's easier to run the "pipeline"


data-pipeline/README.md
```diff
@@ -72,6 +72,38 @@ Follow the directions below to set this project up on your local machine.
...
+
+    ```bash
+    .venv\Scripts\activate
+    ```
+
+4. Install the project dependencies. The following commands can also be used to update dependencies after pulling upstream code changes:
+
+    ```bash
+    # if you're planning to run the scripts without making code changes
+    pip install -r requirements/dev-requirements.txt && pip install -e .
```
bsweger, line 91: Looks like a rebase went awry--will add a commit to fix this up.

data-pipeline/src/covid_variant_pipeline/__init__.py
bsweger, line 1: Haven't used [structlog](https://www.structlog.org/en/stable/) before, but I found the setup boilerplate vastly preferable to Python's out-of-the-box logger config.

Best way to review this is probably to run the pipeline on a local machine and see what you think of the logs

data-pipeline/src/covid_variant_pipeline/assign_clades.py
```diff
@@ -123,14 +126,27 @@ def assign_clades(reference_tree: str):
     return assignment_file
 
 
-def main():
-    logger.info("Starting pipeline")
+@click.command()
+@click.option(
+    "--as-of-date",
+    type=click.DateTime(formats=["%Y-%m-%d"]),
```
bsweger, line 132: setting this parameter as DateTime type is just a lazy way of validating the input

data-pipeline/src/covid_variant_pipeline/assign_clades.py
```diff
@@ -123,14 +126,27 @@ def assign_clades(reference_tree: str):
...
+
+    # incoming as_of_date comes in as a datetime object (for validation purposes), convert back to string now
+    as_of_date = as_of_date.strftime("%Y-%m-%d")
+    logger.info("Starting pipeline", as_of_date=as_of_date)
+
     get_sequences()
     get_sequence_metadata()
-    reference_tree_file = get_reference_tree()
-    assignment_file = assign_clades(reference_tree_file)
+    reference_tree_path, root_sequence_path = save_reference_info(as_of_date)
```
bsweger, line 146: Until we're further along, I've been saving the intermediate outputs to disk (though not committing them, though I could add them to the repo if it would be helpful)

data-pipeline/pyproject.toml
```diff
@@ -11,27 +11,34 @@ readme = "README.md"
...
 [project.optional-dependencies]
 dev = [
     "pytest",
+    "rich",
 ]
 
-[project.entry-points.'console_scripts']
-assign_clades = 'covid_variant_pipeline.assign_clades:main'
+[project.entry-points."console_scripts"]
+assign_clades = "covid_variant_pipeline.assign_clades:main"
```
elray1, line 31: this is awesome

data-pipeline/src/covid_variant_pipeline/assign_clades.py
elray1: I support not committing stuff for now.  We need to sort out where everything will live, and at the moment I don't think we're creating any outputs we'll want/need in the long term.

elray1 (approved):
I filed a couple of issues that came out of playing around with this that it didn't seem necessary to solve as part of this PR.  Thanks!

PR #9, opened 2024-05-13
Bsweger/merge clade assignments and metadata

### LINKED ISSUE #2: functionality to merge metadata and clade assignment outputs

bsweger:
resolves #2

Generate an output file that merges clade assignments and and sequence metadata.

The body of the code is getting pretty ugly, passing all the filenames around. Once we get definitive feedback about this process, we can worry about tidying things up (and making the code base production-ready, which is certainly is not right now)


data-pipeline/src/covid_variant_pipeline/assign_clades.py
```diff
@@ -129,14 +187,26 @@ def main(as_of_date: str):
...
-    get_sequences()
-    get_sequence_metadata()
+    sequence_dir = get_sequences(run_time)
+    metadata_file = get_sequence_metadata(run_time, sequence_dir)
     reference_tree_path, root_sequence_path = save_reference_info(as_of_date)
-    assignment_file = assign_clades(as_of_date, reference_tree_path, root_sequence_path)
+    assignment_file = assign_clades(run_time, sequence_dir, reference_tree_path, root_sequence_path)
 
-    logger.info("Sequence clade assignments are ready", assignment_file=assignment_file)
+    merged_data = merge_metadata(as_of_date, now, metadata_file, assignment_file)
```
bsweger, line 200: This is returning every column available in the NCBI metadata and the output of the assignment CLI (see any *_clade_assignments.csv file in the data/assignment folder).

If it's useful, we can drop whatever columns aren't needed (or supply what we want in a list somewhere and use it to filter)

data-pipeline/src/covid_variant_pipeline/data/assignment/20240513T142649_clade_assignments.csv
bsweger, line 1: Included a few of the output files as examples

data-pipeline/src/covid_variant_pipeline/assign_clades.py
```diff
@@ -1,30 +1,40 @@
+import datetime
 import json
+import os
 import subprocess
 from importlib import resources
 
+import polars as pl
```
bsweger, line 7: we could easily just do this in pandas, but I admit to being polars-curious, so I threw it in here as we're experimenting (https://pola.rs/)

data-pipeline/src/covid_variant_pipeline/assign_clades.py
```diff
@@ -116,6 +137,43 @@ def assign_clades(as_of_date: str, reference_tree: AnyPath, root_sequence: AnyPa
...
+    df_assignments = parse_sequence_assignments(df_assignments)
+
+    joined = df_metadata.join(df_assignments, left_on="Accession", right_on="seq", how="left")
+    joined = joined.with_columns(
+        sequence_retrieved_datetime=run_datetime,
+        reference_tree_date=datetime.datetime.strptime(as_of_date, "%Y-%m-%d").date(),
+    )
+    num_sequences = joined.shape[0]
+
+    missing_clade_assignments = joined.filter(pl.col("Accession").is_null())
```
elray1, line 162: I have two questions on this part (that I'll ask in three pieces):

1. What do we know about the metadata file and the sequence file:
    1. do they always have the same number of entries, for the same set of sequences?
    2. might the metadata file have stuff that's not in the sequence file?
    3. might the sequence file have stuff that's not in the metadata file?
2. Question 1 was a preliminary for understanding why we're doing a left join of assignments into metadata.  That seems safe if we know that we're not in case 1.iii.  Otherwise, a right join or outer join might be safer?
3. Not having dug into how polars works, my interpretation of what's happening in line 162 is that we're keeping rows of the joined df where the Accession is missing.  Based on the left join operation, I would have expected that the filter operation we'd want would be for rows that didn't end up being assigned a nextstrain clade, i.e. something like `pl.col("clade_nextstrain").is_null()`

data-pipeline/tests/unit/util/test_sequence.py
```diff
@@ -0,0 +1,36 @@
...
+from covid_variant_pipeline.util.sequence import parse_sequence_assignments
+
+
+@pytest.fixture
+def df_assignments():
+    return pl.DataFrame(
+        {
+            "seqName": [
+                "PP782799.1 Severe acute respiratory syndrome coronavirus 2 isolate SARS-CoV-2/human/USA/NY-PV74597/2022",
+                "ABCDEFG Severe caffeine deprivation virus",
```
elray1, line 14: 😂

elray1 (commented):
I had one (set of) question(s) about how we're merging metadata and clade assignments, and what missing values might look like.

data-pipeline/src/covid_variant_pipeline/assign_clades.py
bsweger: good questions...will tackle in separate comments, starting with the first

For this iteration, I was working under the assumption that because the sequence data and the sequence metadata are part of the same NCBI "package" that they would contain the same number of entries.

Using [this method for determining the number of sequence in a fasta file,](https://bioinformaticsworkbook.org/Appendix/Unix/unix-basics-3grep.html#gsc.tab=0) and using the files I downloaded in a pipeline run timestamped `20240513T142649`, we can see this is true:

```python
import os
import polars as pl

t = '20240513T142649'
df_metadata = pl.read_csv(f'data/sequence/{t}/{t}-metadata.tsv', separator='\t')
fastafasta = f'data/sequence/{t}/ncbi_dataset/data/genomic.fna'

df_metadata.shape
# (6026, 19)

num_sequences = os.system(f'grep -c "^>" {fastafasta}')
# 6026
```

I love your question because it's never good to assume anything, really. Might the sequences and their metadata differ? They might, and we should check for that.

What do you think we should do if the numbers don't match?

1. Stop the pipeline with an error (i.e., don't continue with the clade assignment process)
2. Continue the pipeline with a warning
3. ¯\_(ツ)_/¯

My sense is that we'd care more about sequences that don't have metadata than the opposite?

FWIW, we're not currently parsing the sequence data directly (i.e., there's no nicely formatted dataframe that represents the fasta file), so we'd need to do that if want to report on specific sequences that don't have a match.

bsweger: For question 2, the prior assumption is indeed the reason for the left join.

Assuming that the metadata file contains all sequences that we care about, we want to maintain them in the final file, even if they don't have a clade assignment.

bsweger: Question 3 -- you are right, and that's a great catch! I'll make a commit to fix this....appreciate the close read.

elray1: thanks!

question 1: i agree, it seems like we care more about sequences without metadata than metadata without sequences.  maybe if there were sequences without metadata we could continue with a warning?  that would let humans look into it to see how serious the issue is afterwards, with results to work with in case they're ok.

question 2: makes sense

question 3: great!

should we file a separate issue for 1/2?

data-pipeline/src/covid_variant_pipeline/assign_clades.py
```diff
@@ -159,11 +159,12 @@ def merge_metadata(
     )
     num_sequences = joined.shape[0]
 
-    missing_clade_assignments = joined.filter(pl.col("Accession").is_null())
+    # ?? what is the difference between "clade" and "clade_nextstrain" ??
```
elray1, line 162: i don't know

elray1 (approved):

data-pipeline/src/covid_variant_pipeline/assign_clades.py
bsweger: https://github.com/reichlab/variant-nowcast-hub/issues/14

PR #12, opened 2024-05-16
WIP Bsweger/update config

bsweger:


hub-config/admin.json
```diff
@@ -8,5 +8,13 @@
     },
     "repository_url": "https://github.com/reichlab/variant-nowcast-hub",
     "file_format": ["parquet"],
-    "timezone": "US/Eastern"
+    "timezone": "US/Eastern",
+    "cloud": {
+        "enabled": false,
```
nickreich, line 13: confirming that we want the cloud set-up here but we have the enabled switch purposely at "false" for now so nothing gets sycned yet.

hub-config/tasks.json
```diff
@@ -8,17 +8,165 @@
...
+                                "type": "double",
+                                "minimum": 0,
+                                "maximum": 1
+                            }
+                        },
+                        "sample": {
+                            "output_type_id_params": {
+                                "is_required": false,
+                                "type": "character",
+                                "max_length": 500,
```
nickreich, line 135: I'd vote for a shorter max length - maybe 15 or 25 or something like that. how much info do we want to allow people to encode here?

hub-config/tasks.json
```diff
@@ -8,17 +8,165 @@
...
+                                "min_samples_per_task": 96,
```
nickreich, line 136: There has been discussion about this elsewhere, but I would advocate for this being set to 100, just to force people to generate 100 samples.

hub-config/tasks.json
```diff
@@ -8,17 +8,165 @@
...
+                                "type": "double",
+                                "minimum": 0,
+                                "maximum": 1
+                            }
+                        }
+                    },
+                    "target_metadata":[
+                        {
+                            "target_id": "variant prop",
+                            "target_name": "variant proportion of all sequences",
```
nickreich, line 150: should be something like "frequency of observed variant lineage". the "of all sequences" didn't sound right to me. But we might also want to be more choosy about the "variant lineage" terminology here.

hub-config/admin.json
bsweger: Yep--wanted to signal our intent to have this on the cloud, but we haven't yet provisioned any of the infrastructure (and I didn't want to make assumptions about what the name of the S3 bucket should be).

hub-config/tasks.json
```diff
@@ -8,17 +8,165 @@
...
+                                "maximum": 1
+                            }
+                        }
+                    },
+                    "target_metadata":[
+                        {
+                            "target_id": "variant prop",
+                            "target_name": "frequency of observed variant lineage",
+                            "target_units": "percentage",
+                            "target_keys": {
```
bsweger, line 152: @nickreich I'm having a tough time parsing this explanation of `target_keys`:

> Should be either null, in the case where the target is not specified as a task_id and is specified solely through the target_id target_metadata property or an object with one or more properties, the names of which match task_id variable(s) named within the same model_tasks object. Each property should have one specified value. Each value, or the combination of values if multiple keys are specified, define a single target value.

Do we need to specify target_keys here, if we're expecting a `variant prop` value that reflects the `compound_taskid_set`?

nickreich: I am not feeling 100% confident in this, but I think we could leave this to null and think about the target as implied by the contents of the row and target id.

bsweger: thanks--just pushed a commit with the next iteration of target_metadata

Issue #7, opened 2024-05-07
pull run settings out into a config file and/or allow them to be specified as command line arguments


elray1:
Looking at [these lines](https://github.com/reichlab/variant-nowcast-hub/blob/330080f85a081f11454ed79b9b1444b335d6d05f/data-pipeline/src/covid_variant_pipeline/assign_clades.py#L13-L21).

I think this would also address an issue where I was asked repeatedly if i wanted to overwrite or rename files when i downloaded some data.  I wasn't sure what path the file names i entered would be relative to, but i think it was one of these things.

bsweger:
Addressed as part of this PR: This change was merged into the `variant-data-pipeline` as part of https://github.com/reichlab/variant-nowcast-hub/pull/15

(there's a new option for specifying the data directory)

PR #15, opened 2024-05-24
Use released_after instead of updated_after when downloading genome package

### LINKED ISSUE #3: Consider changing sequence data filter for initial data pull from `--updated-after` to `--released-after`
I am not sure exactly what the `--updated-after` filter does.  It's probably worth double checking that it reliably gets us everything we need.  It may be that using `--released-after` will get us to a safer/more inclusive data set that definitely includes all sequences with a collection date on or after the specified date.

bsweger:
Resolves #3  #7

This got out of hand, and I broke my own preference for not resolving multiple tickets in a single PR. So apologize for the size (though a lot of the changed files are deletions of old output/intermediary files).

Reviewing commit-by-commit won't tell an especially compelling story, so I don't recommend it. Essentially, these changes entail:

- Adding another parameter to the CLI for specifying "released since date" when grabbing the NCBI genome package 
- For the genome download, use the API instead of the `downloads` CLI (this gives us better control over retries and error handling)
- Streamline the location of the intermediary files and stash that information into a config object. I don't _love_ it, but it's an improvement over the ugly block of global variables


data-pipeline/.pre-commit-config.yaml
```diff
@@ -1,3 +1,8 @@
+exclude: >
```
bsweger, line 1: Stop running the spell checker on genome stuff 😅

elray1 (approved):

PR #19, opened 2024-07-17
Dockerize the data pipeline

### LINKED ISSUE #16: Dockerize the variant data pipeline
For a few reasons, I propose that we "dockerize" our data pipeline sooner rather than later:

- Simplify the setup instructions (users wouldn't have to install various CLIs to make the pipeline work and deal with granting executions privileges to the unsigned software)
- Make it easier to run pipeline on other machines
- Sidestep potential cross-platform pathname issues when writing intermediate data files

What do you think, @elray1?

bsweger:
Resolves #16 

This series of commits:

- adds a Dockerfile to Python data pipeline application
- provides instructions for building the Docker image and using it to run the pipeline
- builds the Docker image to use in GitHub actions
- removes the CLI binaries from the repo, as they're now baked into the Docker image

These can be reviewed commit by commit, to see smaller units of work.


data-pipeline/Dockerfile
```diff
@@ -12,6 +18,7 @@ ARG USER_UID=2222
 ARG USER_GID=$USER_UID
 RUN groupadd --gid $USER_GID $USERNAME \
     && useradd --uid $USER_UID --gid $USER_GID -m $USERNAME
+RUN chown -R pipeline-user:pipeline-user /opt/
```
bsweger, line 21: Added this because running `pytest` and `ruff check` in the container throw errors when the user doesn't have access to write cache files.

data-pipeline/Dockerfile
```diff
@@ -28,7 +35,9 @@ RUN pip install uv
 
 COPY pyproject.toml .
 COPY requirements/ ./requirements/
-RUN uv pip install --system -r ./requirements/requirements.txt --no-cache-dir
+# if sufficiently motivated, we could be clever enough to keep the image size down by
```
bsweger, line 38: I don't love this--GitHub actions provides a CI env variable that could, in theory, be used as a way to only include test files and dependencies when we're building docker in a CI context. But the test assets are lightweight enough that I didn't bother doing that.

data-pipeline/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+ENV PATH "${PATH}:/opt/src/covid_variant_pipeline/bin/"
+
+# Install Python dependencies
+WORKDIR /opt/
+RUN pip install --upgrade pip
+RUN pip install uv
+
+COPY pyproject.toml .
+COPY requirements/ ./requirements/
+RUN uv pip install --system -r ./requirements/requirements.txt --no-cache-dir
```
bsweger: When using `uv` to install, the `--system` flag is required when not installing into a virtual environment.
https://github.com/astral-sh/uv/pull/2046

(Weird imo, but `uv` cuts the install time by 50% or more, so worth it)

.github/workflows/run-integration-tests.yaml
```diff
@@ -0,0 +1,27 @@
...
+        working-directory: data-pipeline
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v4
+
+      - name: Build Docker image 🐳
+        run: |
+          docker build -t data-pipeline .
+
+      - name: run integration tests 🧪
```
bsweger, line 24: To see if this would work (specifically, was unsure how the CLI tools would fare), I temporarily added an integration test step to the existing "run-code-checks.yaml".

Results here: https://github.com/reichlab/variant-nowcast-hub/actions/runs/9980645497/job/27582521804

data-pipeline/Dockerfile
elray1: it seems good to always have these anyways, right? then we can dev against the docker image?

data-pipeline/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+COPY pyproject.toml .
+COPY requirements/ ./requirements/
+# if sufficiently motivated, we could be clever enough to keep the image size down by
+# only installing the dev requirements and test scripts during CI/CD builds
+RUN uv pip install --system -r ./requirements/requirements-dev.txt --no-cache-dir
+RUN uv pip install --system -e .
+
+# Override polars install to use Polars version compatible with legacy CPUs
+# TODO: consider removing the Polars dependency altogether?
```
elray1, line 44: May be worth holding off on removing polars until we think through what we're doing about the other data munging scripts for this hub. I believe at least one of them is using polars, and if we decide it's advantageous to somehow bundle those scripts in the same place as this work it could be nice to have this around.

data-pipeline/Dockerfile
bsweger: Agree, I'm over-thinking it because these dependencies likely don't increase the size of the Docker image enough to really worry about it.

In instances where the size does matter, you'll often see a pattern where the image is built w/o dev dependencies by default. Then for local dev and CI runs, the docker `build` command is passed a flag that instructs docker to use `requirements-dev` instead.

data-pipeline/README.md
```diff
@@ -13,20 +13,56 @@ This is _not_ production code. The purpose of the code is to help us see if the
...
+
+3. Run the pipeline, passing in required arguments:
+
+    ```bash
+    docker run --platform linux/amd64 \
+    -v ./data:/home/pipeline-user/covid_variant \
+    variant-nowcast-hub assign_clades \
+    --sequence-released-since-date 2024-07-16 \
+    --reference-tree-date 2024-07-16 \
+    --data-dir /home/pipeline-user
```
elray1, line 45: With this command, I got the error `Got unexpected extra argument (assign_clades)`, my understanding is because `assign_clades` was specified as the `ENTRYPOINT` in the Dockerfile and so anything that gets specified here is being passed as an argument to `assign_clades`.  It worked for me to just remove that:
```suggestion
    docker run --platform linux/amd64 \
    -v ./data:/home/pipeline-user/covid_variant \
    variant-nowcast-hub \
    --sequence-released-since-date 2024-07-16 \
    --reference-tree-date 2024-07-16 \
    --data-dir /home/pipeline-user
```

However, I don't see the data showing up anywhere!

elray1 (changes requested):
Looks good overall!  However, it seems like a minor tweak to the `docker run` command suggested in the README is needed, and I didn't see output data showing up when I ran that.

PR #39, opened 2024-09-13
[WIP] draft of guidelines as main README.

nickreich:


README.md
```diff
@@ -1,27 +1,87 @@
...
 
+This clade selection is based on the [NextStrain USA sequence count file based on US data from GenBank](https://data.nextstrain.org/files/workflows/forecasts-ncov/open/nextstrain_clades/usa.tsv.gz), using [this script](https://github.com/reichlab/virus-clade-utils/blob/main/src/virus_clade_utils/get_clade_list.py). The NextStrain file with counts is [typically updated daily in the late evening US eastern time](https://github.com/nextstrain/forecasts-ncov/actions/workflows/update-ncov-open-clade-counts.yaml) (it is only updated when new data are available). The hub pulls the most recent version of the file when the workflow runs each week. The precise lineage assignment model (sometimes referred to as a “reference tree”) that was used as well as the version of raw sequence data is be stored as metadata, to facilitate reproducibility and evaluation. 
+<!-- TODO: Check this -->Versions of these target data files are maintained by the hub, so modeling teams can come back and retrospectively test their models on data as they were available in real time.
```
bsweger, line 26: I'm not quite sure what we want to save here: the sequence metadata we used to generate the clade list? The reference trees? Both?

@elray1 and I discussed saving the reference trees, and I created two related issues:
- https://github.com/reichlab/virus-clade-utils/issues/20
- https://github.com/reichlab/variant-nowcast-hub/issues/41

Happy to chat synchronously about what we want to do here, if that's easier!

README.md
```diff
@@ -1,27 +1,87 @@
...
+<!-- TODO: check the above statement is true - is it the Monday date after which no reported data is considered, or a Wednesday? -->
+One implication of this decision is that different numbers of days may be evaluated for some locations when compared with others.
+
+2. The reference phylogenetic tree that defines clades changes over time. Nowcasts and forecasts will be evaluated against whatever sequence data is available 90 days after the deadline that a set of predictions were submitted for. Additionally, those sequences will be assigned a clade based on the reference tree that was used to generate the list of predicted clades on the Monday prior to the submission date. This means that new clades that emerge in the time since the predictions were made will still be classified as they would have been when predictions were made.
```
bsweger, line 54: > This means that new clades that emerge in the time since the predictions were made will still be classified as they would have been when predictions were made.

Should the beginning of this sentence be _This means that new **sequences** that emerge in the time since the predictions where made_?

README.md
```diff
@@ -1,27 +1,87 @@
...
+
+### Point forecast evaluation
+
+Point predictions $\hat \theta$ will be scored directly using the categorical Brier score, comparing the predicted clade proportions of sequences to the observed clade proportions on a specific day in a specific location [[Susswein et al. 2023](https://www.medrxiv.org/content/10.1101/2023.01.02.23284123v4)]
+
+## Probabilistic forecast evaluation
+
+Since full predictive distributions of clade probabilities are solicited in as samples of a predictive distribution, we aim to evaluate the full predictive distribution using a precise scoring procedure, however the precise details of this evaluation are still being worked out. The working proposal for probabilistic forecast evaluation is as follows.
```
bsweger, line 72: > are solicited in as samples of a

extra `in`?

bsweger (commented):
This is great--excited for the hub to open!

Had a few questions and comments, but nothing serious.

README.md
nickreich: Perhaps we need to add something along the lines of what you are suggesting, but what we are talking about is different here. This sentence (as is) refers to the situation where the reference tree evolves so at the time of evaluation there are new __clades__ that were not present at the time the predictions were submitted. We don't expect modelers to be able to see into the future about which clades will exist in 90 days.

nickreich:
@bsweger Thanks for these comments! I'm going to collect a set of comments on this and then make an update early this coming week.

README.md
bsweger: Others may interpret differently, but my read of the sentence as written is that the [new] clades themselves would be classified as they would have been 90 days ago, which is confusing because they didn't exist 90 days ago. Isn't it the sequences themselves that are being classified [as a specific clade] as they would have been 90 days ago

README.md
```diff
@@ -1,27 +1,88 @@
...
+
+3. The variance in the eventually observed clade counts depends on the eventual sample size, or number of sequences tested on a particular day. With a large number of sequences, the variance of the clade counts would tend to be larger and with a small number of sequences the variance would be smaller. However, the number of sequences itself is not of particular epidemiological interest. The evaluation plan introduced below evaluates the counts assuming they follow a multinomial observation model with sample size equal to the number of sequences collected on the target date and location that have been reported as of the evaluation date, so as to eliminate the nuisance parameter of the count variance.
```
dylanhmorris, line 55: While this is true, it I think it could mislead. Arguably the key point is about the variance of the sample fraction $C_k/N$, not the variance of the bin count $C_k$. The variance of $C_k/N$ becomes smaller, not larger, as $N$ increases.

README.md
nickreich: I see what you mean and agree with your logic. I made the update.

bsweger (approved):
Thanks for the updates!

PR #53, opened 2024-09-23
Add python version instructions to the uv metadata

bsweger:
It turns out that `uv` will not recognize the `src/.python-version` if `uv run` is executed from the hub's root. 

So rather than a .python-version file, we'll embed the required Python version in metadata at the top of `get_clades_to_model.py`


src/get_clades_to_model.py
```diff
@@ -19,6 +19,7 @@
 """
 
 # /// script
+# requires-python = ">=3.11"
```
zkamvar, line 22: Is this some UV-specific incantation that avoids the usual requirements.txt and .python-version?

zkamvar (approved):
It appears to work 

```
$ uv run src/get_clades_to_model.py 
Reading inline script metadata from: src/get_clades_to_model.py
 Updated https://github.com/reichlab/virus-clade-utils/ (d3a5874)
Installed 34 packages in 432ms
2024-09-23T19:52:46.056338Z [info     ] genome metadata downloaded     elapsed=30.139455458000157 filename=sequence.py lineno=84
2024-09-23T19:53:06.161027Z [info     ] generated clade list           elapsed=20.094820833000995 filename=get_clade_list.py lineno=64
```

PR #60, opened 2024-09-25
updating admin config to v3.0.1

nickreich:
this addresses half of #34 

I ran this on the config to ensure that it was valid:
```
> library(hubAdmin)
> validate_config(config="admin")
[1] TRUE
✔ ok:  hub-config/admin.json (via admin-schema v3.0.1)
```

Issue #34, opened 2024-09-12
Finalize admin and model-metadata-schema config files


bsweger:
Before opening this hub, we need to finalize the config files below:

- [ ] [admin.json](https://github.com/reichlab/variant-nowcast-hub/blob/main/hub-config/admin.json)
- [ ] [model-metadata-schema.json](https://github.com/reichlab/variant-nowcast-hub/blob/main/hub-config/model-metadata-schema.json)

PR #62, opened 2024-09-25
updated model metadata schema

nickreich:
This is a draft model metadata schema file. Are there additional fields or instructions we want to include? 

This version passes schema/config validation:
```
> validate_hub_config()
...
── $model-metadata-schema 
[1] TRUE
✔ ok:  hub-config/model-metadata-schema.json (from default json schema )
```

Note that this, along with #60 is related to #34

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ We ask modeling teams to submit predictions of frequencies of the predominant SA
 
 ### Submitting to the hub
 
-**How to submit**: Any team wishing to submit a set of nowcasts to the hub may submit a pull-request to this repository with a valid submission file in parquet format. The submission file must be located in `model-output/<team>-<model>/` (where `<team>` and `<model>` are the abbreviated names of your team and your model, respectively), and the submission file must be named `YYYY-MM-DD-<team>-<model>.parquet` where `YYYY-MM-DD` is the date of the Wednesday on which the submission is due. Teams must also submit a model metadata file (see details [below](#model-metadata)). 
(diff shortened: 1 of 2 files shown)


nickreich:
@dylanhmorris and @kaitejohnson would be curious for your feedback on this as well.

kaitejohnson:
My additional suggestions would be for optional but encouraged fields for:
- link to github repository 
- link to long form model description (e.g. pre-print or markdown file)
-  description of data sources (e.g. NextStrain  + GISAID for sequences outside of the U.S. + wastewater variant proportions)

elray1:
couple of thoughts r.e. Kaitlyn's suggestions (all of which I like):

- could have a more general concept of "code archive", that is, allowing for services other than github.  It might be possible to borrow [this specification](https://github.com/hubverse-org/schemas/blob/2148965efadac7c8199cc670ec84db711fd58f1f/v3.0.1/admin-schema.json#L37-L58) for that.
- just throwing out the idea: would it be too much to actually make the link to a formal model writeup that Kaitlyn suggested a required field?  the downside is that teams may be under a crunch at the start of the season, and this would mean that they couldn't actually submit predictions without having set up a place with their writing.

nickreich:
To address the comments about methods write-ups, in the latest revision I:
- made the `methods` field be a longer (1000 characters) abstract of the methods. (required)
- removed the `methods_long` field (was unlimited)
- added a `methods_url` field (required)

It felt weird to require a brief summary, a longer summary, and then a write-up. So I went for a less-brief summary and then a full write-up, both required.

I also am adding a `data_sources` field.

elray1:
@nickreich are the changes you described just above committed/pushed?

nickreich:
almost

nickreich:
they are now.

elray1 (approved):
lgtm

Issue #61, opened 2024-09-25
set up CI workflows for validating submissions


nickreich:
We should be able to set up all of the standard file-format validations and then also add the custom validations from here: https://github.com/reichlab/variant-nowcast-hub/issues/55

nickreich:
The main validations are in place, now just waiting on the custom ones from #55

nickreich:
Going to close this, as we are tracking this in another issue now. #75

PR #64, opened 2024-09-26
adding CI workflow files

### LINKED ISSUE #63: create CI workflow for validating update to hub config files

nickreich:
The files contained in this PR should initiate the following workflows
1. validate config files
2. validate submission files
3. cache hubvalidation dependencies

The workflow files were installed from here: https://github.com/hubverse-org/hubverse-actions/tree/main

I used hubCI to install the files: https://hubverse-org.github.io/hubCI/

closes #63
addresses part of #61 (not the custom validation part)


nickreich:
I think it will benefit us to have these files merged in sooner rather than later, as it will then force us to ensure that any changes to config files are valid and allow us to test example submission files.

elray1 (approved):
approved

PR #67, opened 2024-09-26
Readme updates

nickreich:
see commit messages for specific minor changes.


README.md
```diff
@@ -1,23 +1,23 @@
...
+**How to submit**: Any team wishing to submit a set of nowcasts to the hub may submit a pull request to this repository with a valid submission file in parquet format. The submission file must be located in `model-output/<team_abbr>-<model_abbr>/` (where `<team_abbr>` and `<model_abbr>` are the abbreviated names of your team and your model, respectively, that match the fields in your model metadata file), and the submission file must be named `YYYY-MM-DD-<team>-<model>.parquet` where `YYYY-MM-DD` is the date of the Wednesday on which the submission is due. Teams must also submit a model metadata file (see details [on the Model Metadata README page](https://github.com/reichlab/variant-nowcast-hub/blob/main/model-metadata/README.md)).
```
elray1, line 14:
```suggestion
**How to submit**: Any team wishing to submit a set of nowcasts to the hub may submit a pull request to this repository with a valid submission file in parquet format. The submission file must be located in `model-output/<team_abbr>-<model_abbr>/` (where `<team_abbr>` and `<model_abbr>` are the abbreviated names of your team and your model, respectively, that match the fields in your model metadata file), and the submission file must be named `YYYY-MM-DD-<team_abbr>-<model_abbr>.parquet` where `YYYY-MM-DD` is the date of the Wednesday on which the submission is due. Teams must also submit a model metadata file (see details [on the Model Metadata README page](https://github.com/reichlab/variant-nowcast-hub/blob/main/model-metadata/README.md)). 
```

README.md
```diff
@@ -32,8 +32,8 @@ As we will describe in further detail below, the target for prediction is the pr
 
 To submit probabilistic predictions, a [sample format](https://hubverse.io/en/latest/user-guide/sample-output-type.html) is used to encode samples from the predictive distribution for each model task. The hub requires exactly 100 samples for each model task. One key advantage to submitting sample-based output is that dependence can be encoded across horizons (corresponding to trajectories of variant prevalence over time), or even across locations (see details in [Hubverse sample model-output specifications](https://hubverse.io/en/latest/user-guide/sample-output-type.html#compound-modeling-tasks)). For this hub, we require that samples be submitted in such a way as to imply that they are structured into trajectories across clades and horizons. (See following section for how variants are classified into clade categories.) This means that
```
elray1, line 33:
```suggestion
To submit probabilistic predictions, a [sample format](https://hubverse.io/en/latest/user-guide/sample-output-type.html) is used to encode samples from the predictive distribution for each model task. The hub requires exactly 100 samples for each model task. One key advantage to submitting sample-based output is that dependence can be encoded across horizons (corresponding to trajectories of variant prevalence over time), or even across locations (see details in [Hubverse sample model-output specifications](https://hubverse.io/en/latest/user-guide/sample-output-type.html#compound-modeling-tasks)). For this hub, we require that samples be submitted in such a way as to imply that they are structured into trajectories across clades and horizons. (See following section for how variants are classified into clade categories.) In particular, a common sample ID will be used in multiple rows of the submission file with different combinations of clade and horizon. This means that 
```

elray1 (commented):
some minor suggestions

nickreich:
Thanks for the suggestions! I accepted both.

nickreich:
@elray1 note that I just pushed some larger changes, including a new section to the README that includes some small example tabular data, so people can see what the files will look like.

@kaitejohnson @elray1 do we have strong feelings about whether the column should be named `variant` or `clade`? same with `nowcast_date` vs. `reference_date`? Now would be the time to change if we don't like what we have.

kaitejohnson:
I think Andy might have strong feelings towards clade, and I believe that is the language I have seen us go with in the guidelines (unless these have changed). I personally don't have a preference other than that we be consistent! If we do have a reason for clade we should probably define a clade somewhere... will send this to Andy,

I think `nowcast_date` is more clear, could also go with `submission_date` (since this is both a nowcast and a forecast) but I think either works. 

Otherwise LGTM!

nickreich:
We are going to switch over to clade from variant. I had seen that too and wondered about making the change.

afmagee42:
Short opinion: 
I think clade is a better choice and free of baggage. I agree with Kaitlyn re: `nowcast_date`/`submission_date` > `reference_date`.

Long opinion:
As far as I can tell, there's a collective sense among many that the nomenclature around discussing what gets "variant nowcasted" has gotten rather muddled. There's variant as in "variant under monitoring" or "variant of interest/concern" and such, but there's also variant as in "the thing we're modeling." I get the sense the latter of these is not a usage that many people are truly happy with, but it works in the sense of variant as a variety.

Depending on how you want to look at it, this mess is either freedom not to worry too much, or a reason to try to avoid it.

If trying to avoid it, clade is entirely appropriate because Nextstrain calls these modeling units clades (other than "other," we are also actually modeling [clades](https://en.wikipedia.org/wiki/Clade) in the usual evolutionary sense of the word too). 

There's also [taxon](https://en.wikipedia.org/wiki/Taxon) which has not yet been burdened by association with any particular taxonomy scheme. And lineage, which has unfortunately been taken to be synonymous with "Pango lineage" by many. For the purposes of "variant" nowcasting as it has been done (tracking evolutionary groupings), any one of clade, lineage, or taxon _should_ be appropriate, technically speaking.

rogersbw (approved):

PR #70, opened 2024-09-30
paramaterize setup-r for create-modeling-round

### LINKED ISSUE #69: Setup R should use installed version of R and the Posit Package Manager to install packages
The workflow to create a new modeling round [currently failing because it needs the `libcurl` C library](https://github.com/reichlab/variant-nowcast-hub/actions/runs/11098686061)

The culprit are on lines 36 and 37:

https://github.com/reichlab/variant-nowcast-hub/blob/0d8949b45830d724d5f15b1af80a5edaa0614939/.github/workflows/create-modeling-round.yaml#L35-L38

We should do two things:

1. use the installed version of R (by setting `install-r` to `false`)
2. use the [Posit Package Manager](https://packagemanager.posit.co/client/#/) (formerly RStudio Package Manager) to provide binaries for Linux.

```diff

      - name: Set up R 📊
        uses: r-lib/actions/setup-r@v2
        with:
          install-r: false
          use-public-rspm: true

```

zkamvar:
- use installed version of R
 - use Posit Package Manger
 - set the hubverse-org as an extra repository.

This will fix #69

### DIFF
--- a/.github/workflows/create-modeling-round.yaml
+++ b/.github/workflows/create-modeling-round.yaml
@@ -35,6 +35,10 @@ jobs:
 
       - name: Set up R 📊
         uses: r-lib/actions/setup-r@v2
+        with:
+          setup-r: false
+          use-rspm: true
+          extra-repositories: 'https://hubverse-org.r-universe.dev'
 
       - name: Set up renv 📦
         uses: r-lib/actions/setup-renv@v2


zkamvar:
Note: I might also be able to parameterize this action to check itself on pull request if you want to test this.

bsweger:
Thanks for this, @zkamvar! The change makes sense based on my reading, but I'm wondering: why did this workflow run successfully last week?

zkamvar:
> why did this workflow run successfully last week?

That's a very good question and the short answer is because the renv file changed in https://github.com/reichlab/variant-nowcast-hub/commit/58281ab33fd063721e04db5879ed89a6ca49fa72. 

This caused the cache to be invalid.

The workflow ran in the past because the repository in the lockfile was set to the PPM, but the last commit changed it to use regular CRAN:

```
+       "URL": "https://packagemanager.posit.co/cran/latest"
-        "URL": "https://cloud.r-project.org"
```

bsweger:
> That's a very good question and the short answer is because the renv file changed in [58281ab](https://github.com/reichlab/variant-nowcast-hub/commit/58281ab33fd063721e04db5879ed89a6ca49fa72).
> 
> This caused the cache to be invalid.
> 

Ah, yes--that's the commit that bumped hubAdmin to the new release. Even so, that was 4 days ago, and the last successful GitHub action run for that branch (prior to merging the PR) was 3 days ago.

I assume the other factor here was this other action (from @nick's updates) started caching some R things? https://github.com/reichlab/variant-nowcast-hub/actions/runs/11079486573/job/30788618758

bsweger (approved):

zkamvar:
> Ah, yes--that's the commit that bumped hubAdmin to the new release. Even so, that was 4 days ago, and the last successful GitHub action run for that branch (prior to merging the PR) was [3 days ago](https://github.com/reichlab/variant-nowcast-hub/actions/runs/11071269861).
> 
> I assume the other factor here was this other action (from @nick's updates) started caching some R things? https://github.com/reichlab/variant-nowcast-hub/actions/runs/11079486573/job/30788618758

The missing bit of information is that the renv cache and the other caches go to different places (which we should probably fix, but it's not that big of an issue).

PR #71, opened 2024-09-30
Update create-modeling-round.yaml

zkamvar:
🤦🏽‍♂️ Ugh, it's `use-public-rspm` not just `use-rspm`

This addresses #69 and fixes a bug introduced in #70


bsweger (approved):
Thanks!

PR #75, opened 2024-09-30
add custom validation function

### LINKED ISSUE #55: Create a custom Hub validation to check the sum of submitted values for a given state/location

zkamvar:
This will create a custom validation function called `clad_prop_sum_one()` which will check that the values grouped by modeling task and sample ID sum to one. 

## Testing

You can test this locally as long as you have the `hubValidations` package installed:

```r
library(hubValidations)
source("src/validations/R/clade_prop_sum_one.R")
identical(test(), NULL) # should return TRUE otherwise, it will error
```

## Demo

This will output an `<error/check_failure>` class object on error and a `<message/check_success>` class object on success. They look like this:

The error looks like this (where the example had two tasks that failed):

```
> clade_prop_sum_one(example_tbl(make_it_bad = TRUE), "test")
<error/check_failure>
Error:
! All clade proportions in `value` columns do not sum to one for each unique
  modeling task.
There were 2 tasks with values that did not sum to one:
| nowcast_date | target_date | location | output_type_id | output_type |
| ------------ | ----------- | -------- | -------------- | ----------- |
| 2024-09-25 | 2024-09-23 | MA | MA01 | sample |
| 2024-09-25 | 2024-09-24 | MA | MA01 | sample |
```

With a single failing task, the grammar will be correct:

```
There was 1 task with values that did not sum to one:
| nowcast_date | target_date | location | output_type_id | output_type |
| ------------ | ----------- | -------- | -------------- | ----------- |
| 2024-09-25 | 2024-09-24 | MA | MA00 | sample |
```

Note I have tried to get the table to have correct widths, but it just won't budge. 

A successful run looks like this: 

```
> clade_prop_sum_one(example_tbl(make_it_bad = FALSE), "test")
<message/check_success>
Message:
All clade proportions in `value` columns sum to one for each unique modeling task.
```

I have not yet tested this in the wider test context and an example file would help immensely. 

This will fix #55


nickreich:
I moved the example data to the model-output/README file, currently on the `readme-updates` branch:
https://github.com/reichlab/variant-nowcast-hub/blob/readme-updates/model-output/README.md

zkamvar:
I have tested this on #87 and have confirmed that it works.

Please let me know if any of the output language needs to be changed, otherwise, I would be happy to walk through it with you if you would like.


```
> hubValidations::validate_submission(".", "UMass-HMLR//2024-10-02-UMass-HMLR.parquet")

── variant-nowcast-hub ────

✔ [valid_config]: All hub config files are valid.

── 2024-10-02-UMass-HMLR.parquet ────

✔ [file_exists]: File exists at path model-output/UMass-HMLR/2024-10-02-UMass-HMLR.parquet.
✔ [file_name]: File name "2024-10-02-UMass-HMLR.parquet" is valid.
✔ [file_location]: File directory name matches `model_id` metadata in file name.
✔ [round_id_valid]: `round_id` is valid.
✔ [file_format]: File is accepted hub format.
✔ [metadata_exists]: Metadata file exists at path model-metadata/UMass-HMLR.yml.
✔ [file_read]: File could be read successfully.
✔ [valid_round_id_col]: `round_id_col` name is valid.
✔ [unique_round_id]: `round_id` column "nowcast_date" contains a single, unique round ID value.
✔ [match_round_id]: All `round_id_col` "nowcast_date" values match submission `round_id` from file name.
✔ [colnames]: Column names are consistent with expected round task IDs and std column names.
✔ [col_types]: Column data types match hub schema.
✔ [valid_vals]: `tbl` contains valid values/value combinations.
✔ [rows_unique]: All combinations of task ID column/`output_type`/`output_type_id` values are unique.
✔ [req_vals]: Required task ID/output type/output type ID combinations all present.
✔ [value_col_valid]: Values in column `value` all valid with respect to modeling task config.
ℹ [value_col_non_desc]: No quantile or cdf output types to check for non-descending values. Check skipped.
ℹ [value_col_sum1]: No pmf output types to check for sum of 1. Check skipped.
✔ [spl_compound_taskid_set]: All samples in a model task conform to single, unique compound task ID set that matches or is coarser than the configured `compound_taksid_set`.
✔ [spl_compound_tid]: Each sample compound task ID contains single, unique value.
✔ [spl_non_compound_tid]: Task ID combinations of non compound task id values consistent across modeling task samples.
✔ [spl_n]: Required samples per compound idx task present.
✔ [clade_prop_sum_one]: All clade proportions in `value` columns sum to 1 for each unique modeling task.
✔ [submission_time]: Submission time is within accepted submission window for round.
```

I have also tested it with a modified version of the submission to confirm that it correctly fails:

```
> hubValidations::validate_submission(".", "sample-model/2024-10-02-sample-model.parquet")

── variant-nowcast-hub ────

✔ [valid_config]: All hub config files are valid.

── 2024-10-02-sample-model.parquet ────

✔ [file_exists]: File exists at path model-output/sample-model/2024-10-02-sample-model.parquet.
✔ [file_name]: File name "2024-10-02-sample-model.parquet" is valid.
✔ [file_location]: File directory name matches `model_id` metadata in file name.
✔ [round_id_valid]: `round_id` is valid.
✔ [file_format]: File is accepted hub format.
✔ [metadata_exists]: Metadata file exists at path model-metadata/sample-model.yml.
✔ [file_read]: File could be read successfully.
✔ [valid_round_id_col]: `round_id_col` name is valid.
✔ [unique_round_id]: `round_id` column "nowcast_date" contains a single, unique round ID value.
✔ [match_round_id]: All `round_id_col` "nowcast_date" values match submission `round_id` from file name.
✔ [colnames]: Column names are consistent with expected round task IDs and std column names.
✔ [col_types]: Column data types match hub schema.
✔ [valid_vals]: `tbl` contains valid values/value combinations.
✔ [rows_unique]: All combinations of task ID column/`output_type`/`output_type_id` values are unique.
✔ [req_vals]: Required task ID/output type/output type ID combinations all present.
✔ [value_col_valid]: Values in column `value` all valid with respect to modeling task config.
ℹ [value_col_non_desc]: No quantile or cdf output types to check for non-descending values. Check skipped.
ℹ [value_col_sum1]: No pmf output types to check for sum of 1. Check skipped.
✔ [spl_compound_taskid_set]: All samples in a model task conform to single, unique compound task ID set that matches or is coarser than the configured `compound_taksid_set`.
✔ [spl_compound_tid]: Each sample compound task ID contains single, unique value.
✔ [spl_non_compound_tid]: Task ID combinations of non compound task id values consistent across modeling task samples.
✔ [spl_n]: Required samples per compound idx task present.
✖ [clade_prop_sum_one]: All clade proportions in `value` columns do not sum to one for each unique modeling task.
There were 2 tasks with values that did not sum to one:
| nowcast_date | target_date | location | output_type_id | output_type |
| ------------ | ----------- | -------- | -------------- | ----------- |
| 2024-10-02 | 2024-09-01 | AL | AL00 | sample |
| 2024-10-02 | 2024-09-01 | AL | AL01 | sample |
✔ [submission_time]: Submission time is within accepted submission window for round.
```

zkamvar:
As I'm thinking about this... I'm wondering if I should modify it so that it reports the number of samples that failed instead of listing each one individually. If an individual sample failed and it's due to a faulty model, then a whole lot more samples will also fail, which would create a huge table. 

What do you think, @nickreich, @bsweger, @IsaacMacarthur?

bsweger:
> As I'm thinking about this... I'm wondering if I should modify it so that it reports the number of samples that failed instead of listing each one individually. If an individual sample failed and it's due to a faulty model, then a whole lot more samples will also fail, which would create a huge table.
> 
> What do you think, @nickreich, @bsweger, @IsaacMacarthur?

Agree with your point about avoiding a huge table and showing the total number of sample failures. Would it be possible to show the total while also including a table that shows the first _n_ tasks that failed (as a kind of "hint" if necessary?).

nickreich:
+1 to Becky's suggestion!

zkamvar:
> Would it be possible to show the total while also including a table that shows the first _n_ tasks that failed (as a kind of "hint" if necessary?).

\*Curly Howard voice\* Certainly!

src/validations/R/clade_prop_sum_one.R
```diff
@@ -0,0 +1,158 @@
...
+    msg_subject = "All clade proportions in {.var value} columns",
+    msg_verbs = c("", "do not"),
+    msg_attribute = "{.emph sum to one} for each unique modeling task.",
+    error_object = error_object,
+    details = details
+  )
+}
+
+
+ALL_ONE <- function(dat) isTRUE(all.equal(sum(dat), 1))
```
nickreich, line 32: We want to add in a numerical tolerance here for being within +/- 0.001 of 1 (this specific tolerance was chosen more or less arbitrarily but it is ["codified" as of now](https://github.com/reichlab/variant-nowcast-hub/blob/bdfac214eb78219fc7c5cc4a32f4594540e78791/model-output/README.md?plain=1#L31)). We could change the tolerance if needed, although I think it's a reasonable balance between pragmatism and precision here.

It is likely that for some specific prediction task calculations the sums would not exactly equal 1 due to machine precision. We don't want it to throw an error in this case. I think the above code does not protect against that necessarily?

In other cases, a model may just not be correctly computing sums, in which case the errors are likely to be larger than 0.001. In these cases we definitely want errors

nickreich (dismissed):
@zkamvar Thanks for figuring out the custom validation stuff under the crunch of a deadline! Overall it looks great. 

I do think we need to address the one comment I added in the review about machine precision and a fixed "tolerance" for the sums (to allow for things to be close to, but not exactly equal, 1). I'm approving the PR because there might be an advantage to just pushing it straight in now as is, given that it largely has the desired functionality and it might be useful to be able to have the functionality in there so we can continue to test the validations. But if we do merge it as is, we'd need to do a fast follow to get the other fix in, as we have run into these machine precision errors frequently in other settings once hubs are live. I'm open to either merging now or trying to get the fix in beforehand. Thoughts?

src/validations/R/clade_prop_sum_one.R
zkamvar: the [`all.equal()` method for numeric](https://rdrr.io/r/base/all.equal.html) does protect against numerical precision errors (with a tolerance set to roughly 1.49e-8), but if the tolerance is expected to be 1e-3, then I can certainly set that.

zkamvar: the deed is done.

zkamvar:
> I'm approving the PR because there might be an advantage to just pushing it straight in now as is, given that it largely has the desired functionality and it might be useful to be able to have the functionality in there so we can continue to test the validations. But if we do merge it as is, we'd need to do a fast follow to get the other fix in, as we have run into these machine precision errors frequently in other settings once hubs are live. I'm open to either merging now or trying to get the fix in beforehand. Thoughts?

It looks like this repo has the setting on where new commits require a new approval! 😅 

I've made the tolerance change so it should be all good to go!

nickreich (approved):
Thanks for the quick response. Looks good to me. I'm going to merge and we can hit it with a test file in a PR to see if it works as expected.

PR #84, opened 2024-10-01
Added UMass Amherst metadata file

rogersbw:


rogersbw:
Added .yml file for UMass Amherst model metadata.

model-metadata/UMass-HMLR.yml
```diff
@@ -0,0 +1,45 @@
+team_name: "UMass Amherst"
+team_abbr: "UMass"
+model_name: "Hierarchical Multinomial Logistic Regression"
+model_abbr: "HMLR"
+model_version: "1.0"
+model_contributors: [
```
nickreich, line 6: I vote for having some method to the madness of how we order ourselves here. how about Isaac first and then alphabetical by last name after that?

nickreich (approved):

PR #88, opened 2024-10-02
Bsweger/update clade list/41

### LINKED ISSUE #41: Update weekly clade list process to save the corresponding Nextstrain metadata_version
Requires completion of  https://github.com/reichlab/virus-clade-utils/issues/20

**Background**

When generating and saving a list of clades to model, we should also get and save information that would be required to get the reference tree in use at the time (for reproducibility).

**Definition of done**

- [ ] [`src/get_clades_to_model.py`
](https://github.com/reichlab/variant-nowcast-hub/blob/main/src/get_clades_to_model.py) requests the information required for reproducibility (see [related issue](https://github.com/reichlab/virus-clade-utils/issues/20))
- [ ] The format of the clade lists in `[variant-nowcast-hub/auxiliary-files/modeled-clades`](https://github.com/reichlab/variant-nowcast-hub/tree/generate_task_json/auxiliary-data/modeled-clades) is changed from a text file to a json file with two high-level keys: `clades` and `meta`. The `meta` section will contain the Nextstrain metadata required to find the reference tree that matches the one used to create the clade list
- [ ] The R script reads the clade list to create a new round is updated to parse the clades from the new json file instead of from a text file

bsweger:
Closes #41 

## Background

- Rather than writing .txt files with clade news to `auxiliary-data/modeled-clades`, write .json (same filename format, just YYYY-MM-DD.json instead of YYYY-MM-DD.txt)
- Add metadata to the .json file that we'll need to find the reference tree in use at the time we generated the clade list
- More information: #41 

## Additional Review Notes

- The PR replaces all former .txt files in `auxiliary-data/modeled-clades` with their updated .json versions
- The `meta` key of the re-generated .json files is empty, since we weren't pull metadata in prior weeks
- Example of the artifacts generated via the GitHub action (this is similar to what next Monday's scheduled run will produce): https://github.com/reichlab/variant-nowcast-hub/pull/89/commits/3a700ae80831ab031c1fc043bed44ed00875849a

### DIFF
--- a/hub-config/tasks.json
+++ b/hub-config/tasks.json
@@ -22,7 +22,7 @@
               "optional": ["AL", "AK", "AZ", "AR", "CA", "CO", "CT", "DE", "DC", "FL", "GA", "HI", "ID", "IL", "IN", "IA", "KS", "KY", "LA", "ME", "MD", "MA", "MI", "MN", "MS", "MO", "MT", "NE", "NV", "NH", "NJ", "NM", "NY", "NC", "ND", "OH", "OK", "OR", "PA", "RI", "SC", "SD", "TN", "TX", "UT", "VT", "VA", "WA", "WV", "WI", "WY", "PR"]
             },
             "clade": {
-              "required": ["24A", "24B", "24C", "other", "recombinant"],
+              "required": ["24A", "24B", "24C", "recombinant", "other"],
               "optional": null
             }
           },
@@ -95,7 +95,7 @@
               "optional": ["AL", "AK", "AZ", "AR", "CA", "CO", "CT", "DE", "DC", "FL", "GA", "HI", "ID", "IL", "IN", "IA", "KS", "KY", "LA", "ME", "MD", "MA", "MI", "MN", "MS", "MO", "MT", "NE", "NV", "NH", "NJ", "NM", "NY", "NC", "ND", "OH", "OK", "OR", "PA", "RI", "SC", "SD", "TN", "TX", "UT", "VT", "VA", "WA", "WV", "WI", "WY", "PR"]
             },
             "clade": {
-              "required": ["24A", "24B", "24C", "other", "recombinant"],
+              "required": ["24A", "24B", "24C", "recombinant", "other"],
               "optional": null
             }
           },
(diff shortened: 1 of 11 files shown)


src/get_clades_to_model.py
```diff
@@ -21,14 +21,17 @@
 # /// script
 # requires-python = ">=3.11"
 # dependencies = [
-#   "virus_clade_utils@git+https://github.com/reichlab/virus-clade-utils/",
+#   "virus_clade_utils@git+https://github.com/reichlab/virus-clade-utils@bsweger/get_nextstrain_ncov_metadata",
```
bsweger, line 24: This is the reason the PR is in draft status...until https://github.com/reichlab/virus-clade-utils/pull/26 is merged, this file is instructing `uv` to install the `virus_clade_utils` feature branch that produces the needed ncov metadata.

hub-config/tasks.json
```diff
@@ -22,7 +22,7 @@
               "optional": ["AL", "AK", "AZ", "AR", "CA", "CO", "CT", "DE", "DC", "FL", "GA", "HI", "ID", "IL", "IN", "IA", "KS", "KY", "LA", "ME", "MD", "MA", "MI", "MN", "MS", "MO", "MT", "NE", "NV", "NH", "NJ", "NM", "NY", "NC", "ND", "OH", "OK", "OR", "PA", "RI", "SC", "SD", "TN", "TX", "UT", "VT", "VA", "WA", "WV", "WI", "WY", "PR"]
             },
             "clade": {
-              "required": ["24A", "24B", "24C", "other", "recombinant"],
+              "required": ["24A", "24B", "24C", "recombinant", "other"],
```
bsweger, line 25: Also fixed up the ordering while in here 😃

src/get_clades_to_model.py
```diff
@@ -21,14 +21,17 @@
 # /// script
 # requires-python = ">=3.11"
 # dependencies = [
-#   "virus_clade_utils@git+https://github.com/reichlab/virus-clade-utils/",
+#   "virus_clade_utils@git+https://github.com/reichlab/virus-clade-utils",
```
bsweger, line 24: (This had been pointing to a feature branch for testing, but said feature has been merged, so switched it back...without the unsightly `/`)

nickreich (approved):
This all looks good to me!

PR #92, opened 2024-10-03
Generate a list of sequence collections by date and location

### LINKED ISSUE #50: Save a list of state/date combinations not to score for a specific variant-nowcast-hub round
**Background**

To address the "scoring fairness" modeling challenge identified in the variant-nowcast-hub's README, this hub's evaluation logic will exclude state/date combinations with reported sequence data as of the reference date for a given round that falls into the bounds of the round's target dates.

In other words, any combination of state and target date (where the target date is within 31 days before the Wednesday reference date of the round) that--as of the 8 PM Wednesday submission deadline--has one or more Sars-COV-2 genome sequences with a collection date equal to that target date will not be scored.

**Definition of done**

- [ ] `virus-clade-utils` can generate a data structure of states and the number of sequences for each collection date between `x` and `y`, where `x` will be 31 days prior to 0 horizon and `y` will be the day of 0 horizon/nowcast date/round_id (both in `YYYY-MM-DD` format).
- [ ] `variant-nowcast-hub` has a new script that gets the above data and  writes it to `auxiliary-data/unscore-location-dates`
- [ ] The above script is scheduled to run on Wednesdays at 8 PM, and the output is a PR with the above data (until we decide we don't need a human review)

bsweger:
~~**DO NOT MERGE UNTIL [virus-clade-utils rename has been merged](https://github.com/reichlab/cladetime/pull/32)**~~ Done!

Closes #50 

WIP so that the people using this information when scoring can take a look at the output. Will add some annotations inline.

Remaining work before the PR comes out of draft:

1. Review and merge [this PR from `virus-clade-utils`](https://github.com/reichlab/virus-clade-utils/pull/27)
2. Respond to feedback about contents and format of the "sequence counts by location and date" file (see inline note)
3. Add the scheduled GitHub action that will run on Wednesdays after round closing


auxiliary-data/unscore-location-dates/2024-10-02.csv
bsweger, line 1: This is the output of the script that counts sequence collections by date and location.

@nickreich would love your thoughts on the notes below. 

* I didn't filter out the zeroes, because it seemed easier for a user to work with a complete dataset than have to infer from missing data which states and days are eligible for scoring.
* The data originates from a Polars DataFrame, so we can write it out as json, parquet, etc. It doesn't have to be a .csv

nickreich: 1. I agree with not filtering out zeroes. And like that we are actually storing the counts themselves as this might allow us later on to compute what fraction of counts were observed at the forecast due date.
2. I don't have a strong feeling about .csv vs. .parquet. I think I might suggest not JSON, for simpler tabular readability into R/python.
3. The more [tidy data way](https://r4ds.had.co.nz/tidy-data.html) to store the data would be to have a column for location, a column for date, and a column for count. Given that all of the evaluation/model-output will work in this way, I suggest reformatting into this "longer" version of the data. I would suggest naming the columns `location`, `target_date` and `count`. So that it could be merged in to a model-output or target-data file at some point down the road as-is. Would making the data "tidy" increase the size of the files? I think even if it would, it might be worth doing.

bsweger: Thanks for the feedback @nickreich. Can definitely put the data into 3rd normal form--it's so small and will always be limited to 31 days by definition, so am not worried about size. I _think_ the current format is what @elray1 and I discussed, so hopefully he can chime in (it's also entirely possible that I'm misremembering things!)

.github/workflows/create-location-date-sequence-counts.yaml
```diff
@@ -0,0 +1,57 @@
+name: Create a file of sequence counts by location and collection date
+
+on:
+  schedule:
+    # Not precise, but GithHub actions don't support time zones
+    - cron: "0 02 * 11-12,1-3 3" # 2 AM UTC every Wed (Nov-Dec, Jan-Mar)
```
bsweger, line 6: Timezones 🙀 

Rounds close at 8 PM US Eastern, and GitHub actions are UTC, so tried to come as close as possible when accounting to daylight savings time (not exact, obviously)

auxiliary-data/unscore-location-dates/2024-10-02.csv
elray1: I think Nick's suggestions make sense.  Very minor -- what do you think about a folder name of `unscored-location-dates`?  (unscored rather than unscore)

elray1 (approved):
Approving this PR, but am noting that we are careful to provide the correct round closing datetime when instantiating a CladeTime object, but as of now I believe this is still being truncated over in CladeTime.  So for now, I believe that our data may still be one day off? (not sure, may depend on timezone conversion...)

bsweger:
> Approving this PR, but am noting that we are careful to provide the correct round closing datetime when instantiating a CladeTime object, but as of now I believe this is still being truncated over in CladeTime. So for now, I believe that our data may still be one day off? (not sure, may depend on timezone conversion...)

Good note! The current state of things is a bit confusing. You're right: cladetime's `download_covid_genome_metadata` [_does_ truncate datetimes that are passed to it](https://github.com/reichlab/cladetime/blob/main/src/cladetime/util/sequence.py#L65).  There's still an open issue to fix that.

However, the code in this PR doesn't use that download method, because it works directly against the sequence metadata on S3 file (i.e., no download required). The function used to get the S3 URL [does use a full timestamp](https://github.com/reichlab/cladetime/blob/main/src/cladetime/util/reference.py#L50).

Now that we're using the CladeTime class as our interface, we'll probably end up making this less confusing by updating the `download_covid_genome_metadata` to have a URL passed to it, instead of having it call `_get_s3_object_url` directly. Thanks for being patient as we make the transition!

PR #94, opened 2024-10-04
Yamlfile

rogersbw:
Addressing metadata file comments

### DIFF
--- a/model-metadata/UMass-HMLR.yml
+++ b/model-metadata/UMass-HMLR.yml
@@ -37,7 +37,7 @@ model_contributors: [
 ]
 license: "CC-BY-4.0"
 team_funding: "This work has been supported by the National Institutes of General Medical Sciences (R35GM119582) and the US Centers for Disease Control and Prevention’s Center for Forecasting and Outbreak Analytics (NU38FT000008). The content is solely the responsibility of the authors and does not necessarily represent the official views of NIGMS, the National Institutes of Health, or CDC."
-methods: "A Bayesian hierarchical multinomial logistic regression model for nowcasting COVID variants. Regression coefficents are modeled hierarchically across variants and locations. Model is adapted from those presented in Abousamra, Figgins and Bedford (2024)."
+methods: "A Bayesian hierarchical multinomial logistic regression model for nowcasting COVID variants. Regression coefficents are modeled hierarchically across variants and locations. Model is adapted from Abousamra, Figgins and Bedford (2024)."
 methods_url: "https://github.com/IsaacMacarthur/Umass-Virus_modeling-code/blob/main/Model%20Details.pdf"
 data_sources: "NextStrain"
 ensemble_of_models: false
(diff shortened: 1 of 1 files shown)


rogersbw:
I imagine we'll want to workshop the methods description a bit more, but I figured this would serve for now.

model-metadata/UMass-HMLR.yml
```diff
@@ -19,27 +24,23 @@ model_contributors: [
...
-methods: "A Bayesian hierarchical multinomial logistic regression model for nowcasting COVID variants. Regression coefficents are modeled hierarchically across variants and locations."
+methods: "A Bayesian hierarchical multinomial logistic regression model for nowcasting COVID variants. Regression coefficents are modeled hierarchically across variants and locations. Model is adapted from Abousamra, Figgins and Bedford (2024)."
 methods_url: "https://github.com/IsaacMacarthur/Umass-Virus_modeling-code/blob/main/Model%20Details.pdf"
 data_sources: "NextStrain"
 ensemble_of_models: false
 ensemble_of_hub_models: false
 repo_url: "https://github.com/IsaacMacarthur/Umass-Virus_modeling-code/tree/main"
+citation: "Abousamra, E., Figgins, M., & Bedford, T. (2024). Fitness models provide accurate short-term forecasts of SARS-CoV-2 variant frequency. PLOS Computational Biology, 20(9), e1012443."
```
nickreich, line 46: I suggest removing this as it wouldn't be quite right to use this citation as an exact citation for our model, which is similar to, but different than, their approach.

model-metadata/UMass-HMLR.yml
```diff
@@ -19,27 +24,23 @@ model_contributors: [
...
-    "email": "maryclaregri@umass.edu"
+    "name": "Benjamin W. Rogers",
+    "affiliation": "Department of Biostatistics and Epidemiology, University of Massachussetts Amherst",
+    "email": "bwrogers@umass.edu"
   },
 ]
 license: "CC-BY-4.0"
 team_funding: "This work has been supported by the National Institutes of General Medical Sciences (R35GM119582) and the US Centers for Disease Control and Prevention’s Center for Forecasting and Outbreak Analytics (NU38FT000008). The content is solely the responsibility of the authors and does not necessarily represent the official views of NIGMS, the National Institutes of Health, or CDC."
-methods: "A Bayesian hierarchical multinomial logistic regression model for nowcasting COVID variants. Regression coefficents are modeled hierarchically across variants and locations."
+methods: "A Bayesian hierarchical multinomial logistic regression model for nowcasting COVID variants. Regression coefficents are modeled hierarchically across variants and locations. Model is adapted from Abousamra, Figgins and Bedford (2024)."
```
nickreich, line 40: Could we say a few ways in which it is similar to/different from their model? E.g. didn't we change the way the hierarchy was used (didn't they do it across variants where we do it across locations?).

nickreich: Maybe also, in place of the citation below, we could add a journal name here, for a bit more of a "complete" citation.

nickreich (changes requested):
Author list looks great. I have a few suggested changes for the citation/reference part.

rogersbw:
Made the suggested edits. As far as I know, the only difference for the current model that Isaac is using is that hierarchies over variants.

nickreich (approved):

Issue #56, opened 2024-09-25
Create example submission files


bsweger:
Create a valid model-output file to serve as an example of what a submission to this hub will look like. Two purposes:

1. test validations (notably, the custom validation we're writing for this hub: https://github.com/reichlab/variant-nowcast-hub/issues/55)
2. show teams what a model-output file should look like

**Definition of done**

There should be two files: one good (where "good" = validates) and one "bad" (where bad = there is at least 1 model task where at least 1 set of samples sums to something other than one)

- [ ] single round (one file, one team)
- [ ] parquet format (this hub accepts parquet only)
- [ ] more than one target date and more than one location, but doesn't need to have them all
- [ ] 100 samples with correct sample ids
- [ ] location: auxiliary-data/example files
- [ ] add description of and link to the example files in the README in the root directory of the hub repo

bsweger:
@nickreich to assign

elray1:
It may be nice to create these examples using some simple, self-contained, and artificial process rather than using an actual prediction model.  For example, you might specify some Dirichlet distributions with arbitrary/hand-selected parameter values to use for each date.

PR #96, opened 2024-10-04
Script to generate simulated data for valid submission

trobacker:
#56 Created an R script that defines a function to make simulated data with default settings for quick use. There's also a sample parquet file ready for use. There's an example validation run at the end of the script.


src/sample_submission_data.R
```diff
@@ -0,0 +1,88 @@
...
+
+sim_model_output <- function(locations = c("AL", "AZ", "CA", "CO", "DE"), 
+                             alpha = c(3,3,2,2,1,1),
+                             test_clades = c("24A", "24B", "24C", "24E", "recombinant", "other"),
+                             target_date = seq(as.Date("2024-09-01"), as.Date("2024-10-12"), by = "1 days"),
+                             nowcast_date = "2024-10-02",
+                             n_samp = 100,
+                             seed = 42
+){
+  ## Simulate data in appropriate model_output format
```
nickreich, line 19: Minor comment, but I think the Roxygen format requires this documentation chunk to be before the function definition.

src/sample_submission_data.R
```diff
@@ -0,0 +1,88 @@
+## Script that generates sample submission data to pass validation checks for
+## the forecasting hub
+## It's not particularly efficient but it's easy to understand
+
+## Function to generate simulated data
+# Uses the Dirichlet distribution to generate random proportions of clades
+require(brms) # for data simulation
```
nickreich, line 7: suggest putting these require statements within the function call. related to my comment below about having one file for the function and one for the script.

src/sample_submission_data.R
```diff
@@ -0,0 +1,88 @@
...
+                   location = loc_col,
+                   output_type = output_type_col,
+                   output_type_id = output_type_id_col,
+                   value = value_col
+  )
+  
+  return(df)
+}
+
+df_sim <- sim_model_output()
```
nickreich, line 82: I suggest having the script be in one file with no stand-alone code, so that the file can be sourced directly in another file that has sources the file and runs these lines of code.

src/sample_submission_data.R
```diff
@@ -0,0 +1,88 @@
...
+  #' @locations Vector of location abbreviations e.g. c("MA", ...)
+  #' @alpha Parameters for rdirichlet for length(test_clades)
+  #' @test_clades Clades for the weekly forecasts
+  #' @target_date Forecast dates
+  #' @nowcast_date Nowcast date
+  #' @n_samp Number of samples per location (100 required,available for testing)
+  #' @seed Set random seed value for RNG
+  
+  set.seed(seed)
+
```
nickreich, line 29: Not necessary to merge this in, but ideally we might have a few basic checks here like:
- length(test_clades) == length(alpha)
- n_samp is a positive integer
- alpha is a numeric vector
- target_date and nowcast_date are in YYYY-MM-DD format

model-output/2024-10-02-umass-validtestsubmission.parquet
nickreich, line 1: The hubverse standards are strict about what files can live within the `model-output` directory. I suggest stashing this file in `auxiliary-data/example-files/`.

nickreich (changes requested):
Thanks for this script and the file! See specific comments (some of which are "optional" to address for now). But the stashing of the example file is a must-have due to the strict storage requirements for hubverse.

model-output/2024-10-02-umass-validtestsubmission.parquet
trobacker: Thanks for feedback! I'll make the changes soon.

nickreich (dismissed):

trobacker (commented):
Looks good! Reviewed with Nick.

bsweger:
@trobacker Added a short note inline, but the longer version is that the R environment in this repo is managed by renv, which means that part of the workflow when introducing new package requirements is to capture them via the `renv::status()` command.

`renv::status()` will update the `renv.lock` file to reflect the new packages (and updates to existing packages). I did that exercise and pushed a commit with the updated renv.lock.

This group (including myself) usually doesn't use `renv` as part of our R workflow, but it's important for ensuring that the hub's scheduled jobs are reproducible.

Pushing the updated renv.lock fixed things on my machine, but to test it yourself:

1. Pull the branch's updates and open the [RStudio project in repo's `src/` directory](https://github.com/reichlab/variant-nowcast-hub/blob/main/src/variant-nowcast-hub.Rproj)
2. From the R console, run `renv::status()`

Hopefully, you'll get a message that the project is in a consistent state and will be able to run your code (once the file paths are fixed up to assume that `src` is the working directory.

src/sample_submission_data.R
```diff
@@ -0,0 +1,9 @@
+## Script that generates sample submission data to pass validation checks
+require(arrow) # for parquet writing
```
bsweger, line 2: here::i_am("src/sample_submission_data.R")
require(arrow) # for parquet writing

bsweger, line 2:
```suggestion
here::i_am("src/sample_submission_data.R")
require(arrow) # for parquet writing
```

src/sample_submission_data.R
```diff
@@ -0,0 +1,9 @@
...
+source('sim_model_output.R')
+# Function found in /src to generate data frame
+df_sim <- sim_model_output()
+
+# Note: make sure to point to the correct path for your data, defaults to folder 'model-output'
+arrow::write_parquet(df_sim, "auxiliary-data/example-files/2024-10-02-umass-validtestsubmission.parquet")
```
bsweger, line 8:
```suggestion
arrow::write_parquet(df_sim, "../auxiliary-data/example-files/2024-10-02-umass-validtestsubmission.parquet")
```

zkamvar, line 8: In conjunction with https://github.com/reichlab/variant-nowcast-hub/pull/96/files#r1793754309, this will make sure the parquet file is always written to the `auxiliary-data/` folder in the root of the hub.


```suggestion
arrow::write_parquet(df_sim, here::here("auxiliary-data/example-files/2024-10-02-umass-validtestsubmission.parquet"))
```

src/sample_submission_data.R
```diff
@@ -0,0 +1,9 @@
+## Script that generates sample submission data to pass validation checks
+require(arrow) # for parquet writing
+source('sim_model_output.R')
```
zkamvar, line 3:
```suggestion
source(here::here('src/sim_model_output.R'))
```

src/sample_submission_data.R
zkamvar: Better to use `here`: https://github.com/reichlab/variant-nowcast-hub/pull/96/files#r1793783070

src/sample_submission_data.R
zkamvar: It is indeed appropriate here (pun intended): 

Information about the package can be found at https://here.r-lib.org

zkamvar: You really want the script to fail here if you don't have arrow. `library()` throws an error if the package is not found while `require()` merely returns `FALSE`.  

```suggestion
library("arrow") # for parquet writing
```

src/sample_submission_data.R
```diff
@@ -0,0 +1,9 @@
...
+# Function found in /src to generate data frame
+df_sim <- sim_model_output()
+
+# Note: make sure to point to the correct path for your data, defaults to folder 'model-output'
+arrow::write_parquet(df_sim, "auxiliary-data/example-files/2024-10-02-umass-validtestsubmission.parquet")
+hubValidations::validate_model_data(".", "2024-10-02-umass-validtestsubmission.parquet")
```
zkamvar, line 9: `here::here()` points to the root of the hub, but I'm not sure this will work exactly since `validate_model_data()` expects your model file to be in the `model-outputs` directory. 

```suggestion
hubValidations::validate_model_data(here::here(), "2024-10-02-umass-validtestsubmission.parquet")
```

src/sample_submission_data.R
trobacker: Cool, thanks for the suggestions! I've not seen the `here::here` before, learn new things every day :)

src/sample_submission_data.R
bsweger: nice--TIL!

src/sample_submission_data.R
nickreich: Does `here::here()` point to the root of the hub or the root of the Rproj directory, which is in `src/`? My workflows of how to set things up have been a bit turned on their heads by the inclusion of the Rproj file in a subdirectory.

nickreich: Looking at the `here` documentation a bit, I've convinced myself that the `here_i_am()` call proposed at the beginning would establish `here::here()` as the `src/` directory. Which might then suggest that we need to replace `here::here()` in the above suggestion with `here::here("../")`, right?

zkamvar: > I've convinced myself that the `here::i_am()` call proposed at the beginning would establish `here::here()` as the `src/` directory.

Nope. The `here::i_am("src/sample_submission_data.R")` statement is saying "This is a file whose path is `src/sample_submission_data.R`, _relative to the root of the project_, thus implying the root of the project. 

You can see this in [`src/make_round_config.R#L11`](https://github.com/reichlab/variant-nowcast-hub/blob/a9bc30382c29740f3447499acc64743e083aaf37/src/make_round_config.R#L11) defining `here::i_am("src/make_round_config.R")` and then [`src/make_round_config.R#L11`](https://github.com/reichlab/variant-nowcast-hub/blob/a9bc30382c29740f3447499acc64743e083aaf37/src/make_round_config.R#L203) defining `hub_root <- here::here()`

nickreich:
I accepted some of the suggested commits above.

trobacker:
> @trobacker Added a short note inline, but the longer version is that the R environment in this repo is managed by renv, which means that part of the workflow when introducing new package requirements is to capture them via the `renv::status()` command.
> 
> `renv::status()` will update the `renv.lock` file to reflect the new packages (and updates to existing packages). I did that exercise and pushed a commit with the updated renv.lock.
> 
> This group (including myself) usually doesn't use `renv` as part of our R workflow, but it's important for ensuring that the hub's scheduled jobs are reproducible.
> 
> Pushing the updated renv.lock fixed things on my machine, but to test it yourself:
> 
> 1. Pull the branch's updates and open the [RStudio project in repo's `src/` directory](https://github.com/reichlab/variant-nowcast-hub/blob/main/src/variant-nowcast-hub.Rproj)
> 2. From the R console, run `renv::status()`
> 
> Hopefully, you'll get a message that the project is in a consistent state and will be able to run your code (once the file paths are fixed up to assume that `src` is the working directory.

Thanks for this explanation, Becky!

src/sim_model_output.R
```diff
@@ -0,0 +1,79 @@
...
+  loc_col = c()
+  clade_col = c()
+  value_col = c()
+  
+  # Create observations using dirichlet distribution for proportions
+  for(loc in locations){
+    for(date in target_date){
+      for(sample_pt in 0:(n_samp-1)) {
+        # Create random proportions with prob alpha for each sample point
+        props <- rdirichlet(1, alpha = alpha)
```
zkamvar, line 41: This modifies the function in two ways:

1. use `withr::local_seed()` to avoid downstream effects of internally setting the seed (it future-proofs the function call so that you can move it around without affecting other random seeds). Since withr is already part of the renv file, this should work as advertised. 
2. uses the `pkg::fn()` syntax to avoid attaching the package into the global namespace.

```suggestion
  # set the `seed` and then resets it to the state of the RNG before this ran
  # https://withr.r-lib.org/reference/with_seed.html
  withr::local_seed(seed)
  
  # more QA to come e.g. n_samp positive integer, etc.
  
  # Columns to be created for output
  target_date_col = c()
  output_type_id_col = c()
  loc_col = c()
  clade_col = c()
  value_col = c()
  
  # Create observations using dirichlet distribution for proportions
  for(loc in locations){
    for(date in target_date){
      for(sample_pt in 0:(n_samp-1)) {
        # Create random proportions with prob alpha for each sample point
        props <- brms::rdirichlet(1, alpha = alpha)
```

src/sample_submission_data.R
```diff
@@ -0,0 +1,9 @@
+## Script that generates sample submission data to pass validation checks
+library("arrow") # for parquet writing
+source(here::here('src/sim_model_output.R'))
+# Function found in /src to generate data frame
+df_sim <- sim_model_output()
+
+# Note: make sure to point to the correct path for your data, defaults to folder 'model-output'
+arrow::write_parquet(df_sim, here::here("auxiliary-data/example-files/2024-10-02-umass-validtestsubmission.parquet"))
+hubValidations::validate_model_data(here::here(), "2024-10-02-umass-validtestsubmission.parquet")
```
zkamvar, line 9: If I understand this correctly, whoever uses this script should manually change `auxiliary-data/example-files/` to `model-outputs/umass-validtestsubmission/`?

zkamvar (changes requested):
There's still one more thing that needs to be done for this to work: the `here::i_am()` call needs to be placed at the top of the R script.

trobacker (commented):
I believe this is good-to-go?

zkamvar (approved):
This is good to go! There is one unresolved question that I had (https://github.com/reichlab/variant-nowcast-hub/pull/96#discussion_r1797300855), but that's not a blocker.

src/sample_submission_data.R
trobacker: If the user wants to validate the data locally via hubValidations I believe that is correct. 
When submitting to the hub, the submission file also needs to to be in their `model-output` team folder ([Hubverse model-output](https://hubverse.io/en/latest/user-guide/model-output.html)).
As this script generates a sample submission file for reference, it seems reasonable to keep it out of such a place and in auxiliary data.

zkamvar: That's what I was assuming. It was the impetus for https://github.com/hubverse-org/hubValidations/issues/130.

Issue #80, opened 2024-10-01
test model submission that does not sum to 1


nickreich:
This can be completed by submitting a file that is fully correct but has at least one value that makes a sum of clade proportions to not 1.

trobacker:
For reference, here's what an example output of this error may look like. In this case, many proportions do not sum to 1, see `[value_col_valid]` below:

```
hubValidations::validate_model_data(hub_path = ".", 
+                                     file_path = "2024-10-02-umass-badtestsubmission.parquet")

── 2024-10-02-umass-badtestsubmission.parquet ────

✔ [file_read]: File could be read successfully.
✔ [valid_round_id_col]: `round_id_col` name is valid.
✔ [unique_round_id]: `round_id` column "nowcast_date" contains a single, unique round ID value.
✔ [match_round_id]: All `round_id_col` "nowcast_date" values match submission `round_id` from file
  name.
✔ [colnames]: Column names are consistent with expected round task IDs and std column names.
✔ [col_types]: Column data types match hub schema.
✔ [valid_vals]: `tbl` contains valid values/value combinations.
✔ [rows_unique]: All combinations of task ID column/`output_type`/`output_type_id` values are
  unique.
✔ [req_vals]: Required task ID/output type/output type ID combinations all present.
✖ [value_col_valid]: Values in column `value` are not all valid with respect to modeling task
  config.  Values 1.64082051954864, 1.39419327558475, 1.7781695793825, 1.56652143950033,
  1.11891236899438, 2.35824530656231, 1.77210188717548, 1.02784946334966, 1.92670602467648,
  1.4257758888811, 2.32423973426083, 1.24590460946878, 1.21852336365822, 2.14001639972308,
  1.0208080346671, 1.11427194788356, 1.08532523128854, 1.16144359215991, …, 1.27441130988298, and
  1.36619765625589 are greater than allowed maximum value 1 for output type "sample".
ℹ [value_col_non_desc]: No quantile or cdf output types to check for non-descending values. Check
  skipped.
ℹ [value_col_sum1]: No pmf output types to check for sum of 1. Check skipped.
✔ [spl_compound_taskid_set]: All samples in a model task conform to single, unique compound task ID
  set that matches or is coarser than the configured `compound_taksid_set`.
✔ [spl_compound_tid]: Each sample compound task ID contains single, unique value.
✔ [spl_non_compound_tid]: Task ID combinations of non compound task id values consistent across
  modeling task samples.
[...]

PR #97, opened 2024-10-06
Invalid submission sum1

trobacker:
#80 Sample model-output submission file that has an entry that _does not sum to 1._


src/sample_submission_data.R
nickreich, line 1: I would suggest removing this script from this PR and just keeping it in #96 , otherwise we will have merge conflicts.

model-output/2024-10-02-umass-validtestsubmission.parquet
nickreich, line 1: same comment as for the script - I recommend having this file in just one PR.

model-output/2024-10-02-umass-invalidsum1submission.parquet
nickreich, line 1: To appropriately test the "sum-to-1" validation, you should rename this file as if it were the UMass-HMLR submission file for last week and put it in the model-output/UMass-HMLR folder. Then validations should run on it as if it were a real submission.

nickreich: As it stands, the validations do not appear to be conducting all of the content validations on the file, I think due in part to it being mis-named and mis-located.

nickreich (changes requested):
See specific comments about only submitting the one invalid file and renaming/re-locating it.

nickreich:
This failed both on submission date and not all sample IDs summing to 1 (both of which were expected. Yay!

Issue #81, opened 2024-10-01
test that validation throws error if not 100 samples


nickreich:
This can be completed by submitting a file that is fully correct but has at least one instance where there are not 100 unique samples.

nickreich:
Closed per investigations in #102

PR #102, opened 2024-10-09
Invalid submission non-100 samples

trobacker:
#81 Example submission file that does not validate by _not_ containing 100 samples for one particular sample. 
Do not merge.


nickreich:
The validation failure here contains this message:
```
ⓧ [spl_non_compound_tid]: Task ID combinations of non compound task id values not consistent across
  modeling task samples.  Sample "AL01" does not match most frequent non compound task ID
  combination for its modeling task. See `errors` attribute for details.
```
I was expecting a clearer output message about there not being 100 samples, but maybe this output makes sense given that not all of the AL01 samples are missing, but maybe just one day is. The code and output below show that the AL01 sample ID is present but has fewer instances than other output types (246 vs 252 - looks like maybe one horizon is missing):
```
> dat <- read_parquet("model-output/UMass-HMLR/2024-10-02-UMass-HMLR.parquet")
> table(dat$output_type_id)
AL00 AL01 AL02 AL03 AL04 AL05 AL06 AL07 AL08 AL09 AL10 AL11 AL12 AL13 AL14 AL15 AL16 AL17 AL18 AL19 
 252  246  252  252  252  252  252  252  252  252  252  252  252  252  252  252  252  252  252  252 
```
So I think this is an appropriate failure.

Noting that when I then removed the rest of the AL01 rows and validated again locally, the file now passed the `[spl_non_compound_tid]` check but failed the `[spl_n]` check:
```
✔ [spl_non_compound_tid]: Task ID combinations of non compound task id values consistent across
  modeling task samples.
✖ [spl_n]: Number of samples per compound idx not consistent.  Sample numbers supplied per compound
  idx vary between 99 and 100.  See `errors` attribute for details.
```

Once again, I think expected and appropriate validation failures. So I'm going to call this closed.

Issue #82, opened 2024-10-01
test that validation throws error if incorrect single value is supplied


nickreich:
This can be completed by submitting a file that is fully correct but has at least one model task where the values sum to 1 but one of the values is outside of [0,1].

PR #103, opened 2024-10-09
Invalid submission incorrect single value supplied

trobacker:
#82 Example submission file that does not validate correctly due to an incorrect single value outside of [0,1] but the sample values sum to 1. In particular, I used values c(-0.1, 0.3, 0.2, 0.2, 0.2, 0.2).


nickreich:
Looks like this failed as expected and desired. Thanks! 
image

PR #110, opened 2024-10-10
Update job name for create-location-date-sequence-counts.yaml

elray1:


nickreich (approved):
LGTM. This aligns the job name with the workflow.

bsweger:
Good catch--thanks!

Issue #104, opened 2024-10-09
test that validation passes when a submission includes both mean and sample predictions


elray1:
3 possible cases:

1. submission file includes only mean predictions
2. submission file includes only sample predictions
3. submission file includes both mean and sample predictions

PR #116, opened 2024-10-16
LANL 10/23 predictions

MKupperman:
LANL Theoretical Biology and Biophysics group predictions using CovTransformer (a transformer model for covid variant predictions) for 10/16 are included in this merge request.

We encountered an issue with "NA" not being correctly recognized as NA when performing parquet validation. Please email me if this is an issue on the back-end.

We have clipped our NA/NaN/Null predictions into 0, to comply with the normalization requirement that frequencies must sum to 1. 

> The views expressed here are those of the authors and does not necessarily represent the official views of the National Institutes of Health, Los Alamos National Laboratory, or the US Government."

PR #117, opened 2024-10-16
Testing a submission with samples and means present.

trobacker:
#104 This PR is to investigate (in)validation of a model submission file with `samples` and `means` present. The file format follows the `model-output` [guidelines](https://hubverse.io/en/latest/user-guide/model-output.html) found on the hubVerse yet doesn't seem to validate.


nickreich:
@annakrystalli reports that the error above is due to similar problems as in #116 .

trobacker:
Note: the submission file does validate but it's important to have the missing values in `output_type_id` as actual missing values: `NA` in R and I assume `np.nan` in Python - not as strings/characters NA. I haven't made a submission via python yet but I imagine writing a pandas data frame with `np.nan` entered to the appropriate format.

MKupperman:
I have tested `pd.NA`, `None`, and `np.nan` to store entries in a parquet with Pandas/PyArrow, through `hubValidations::validate_model_data`. All fail, as the recognized column types are `vctrs_unspecified`, `vctrs_unspecified`, and `double` respectively.

annakrystalli:
Oh interesting. 

Would it be possible to post the full output of `validate_model_data()`

It seems casting to missing values in python is changing or even removing the column datatype. This also happens in R when setting a column's content to all `NA`s which changes the data type to logical (Boolean) by default. In R we can set an all `NA` column to a specific data type though by using a typed version of NA, e.g. To conserve a character column we could assing `NA_character_` instead of `NA`. Is there something equivalent in python?

MKupperman:
Sure! Here's the full output, using `None` in python.

```sh
── 2024-10-23-LANL-CovTransformer.parquet ────



✔ [file_read]: File could be read successfully.
✔ [valid_round_id_col]: `round_id_col` name is valid.
✔ [unique_round_id]: `round_id` column "nowcast_date" contains a single, unique round ID value.
✔ [match_round_id]: All `round_id_col` "nowcast_date" values match submission `round_id` from file name.
✔ [colnames]: Column names are consistent with expected round task IDs and std column names.
✖ [col_types]: Column data types do not match hub schema.  `output_type_id` should be "character" not "vctrs_unspecified".
✔ [valid_vals]: `tbl` contains valid values[/value](http://localhost:8888/value) combinations.
✔ [rows_unique]: All combinations of task ID column[/](http://localhost:8888/)`output_type`[/](http://localhost:8888/)`output_type_id` values are unique.
✔ [req_vals]: Required task ID[/output](http://localhost:8888/output) type[/output](http://localhost:8888/output) type ID combinations all present.
✔ [value_col_valid]: Values in column `value` all valid with respect to modeling task config.
ℹ [value_col_non_desc]: No quantile or cdf output types to check for non-descending values. Check skipped.
ℹ [value_col_sum1]: No pmf output types to check for sum of 1. Check skipped.
ℹ [spl_compound_taskid_set]: No v3 samples found in model output data to check. Skipping `check_tbl_spl_compound_taskid_set` check.
ℹ [spl_compound_tid]: No v3 samples found in model output data to check. Skipping `check_tbl_spl_compound_tid` check.
ℹ [spl_non_compound_tid]: No v3 samples found in model output data to check. Skipping `check_tbl_spl_non_compound_tid` check.
ℹ [spl_n]: No v3 samples found in model output data to check. Skipping `check_tbl_spl_n` check.
✔ [clade_prop_sum_one]: All clade proportions in `value` columns sum to one for each unique modeling task.
```

PR #121, opened 2024-10-23
HMLR 10/23 submission file with samples and means.

trobacker:
UMass HMLR model for 10/23 submission. It includes `output_type` of `sample` and `mean`.


nickreich (changes requested):
There is a space in the filename. This should be fixed.

trobacker (commented):
Removed space.

IsaacMacarthur (approved):

IsaacMacarthur (approved):

nickreich (approved):

nickreich:
Noting that I flagged the whitespace in filename issue as a bug [here](https://github.com/hubverse-org/hubValidations/issues/132).

PR #129, opened 2024-10-29
CADPH-CATaLog, model output 10-30-2024

breesie:
Model output only, only submitting results for CA.


trobacker:
Hello! It looks like you've not included the required team meta data file that the Hub requires. The validation error returned is the following: 
`[metadata_exists]: Metadata file does not exist at path 'model-metadata/CADPH-CATaLog.yml' or 'model-metadata/CADPH-CATaLog.yaml'.`

You can see more details about this team meta data file here:
https://hubverse.io/en/latest/user-guide/model-metadata.html

You can also see a quick summary of the structure on the main README page: 
https://github.com/reichlab/variant-nowcast-hub

Let us know if we can assist further!

PR #130, opened 2024-10-29
Update admin.json to use correct org

zkamvar:


bsweger (approved):
Good catch!

zkamvar:
It was a minor annoying thing because all the hubUtils calls that read in the config were complaining.

PR #132, opened 2024-10-29
Model outputs for CA 10-30-2024, corrected parquet file and corrected yml

breesie:
only submitting preds for CA, changed the NAs in output_type_id to character


trobacker:
Hi @breesie, before or after validation is passed, could you remove the changes to the `variant-hub-modelers.rproj` that have been committed? We should see only your model files for the PR. Thank you!

trobacker:
One more validation issue: 
```
[valid_vals]: `tbl` contains invalid values/value combinations.  Column
  `output_type_id` contains invalid value "NA".
```
You should be able to replace `"NA"` with missing values (e.g. R's `NA`) as opposed to character/strings to get this check passed.

PR #138, opened 2024-11-04
update renv; pin hubVerse packages; pin R

### LINKED ISSUE #137: clade generation script failing because of renv error
[this week's clade generation script run](https://github.com/reichlab/variant-nowcast-hub/actions/runs/11657401248/job/32454916598) is failing because it could not find `hubValidations` 0.7.0.

The reason for this is that we fetched it from the r-universe.... which is non-archival, so once a version changes there, the old version is not accessible from that source. 

The most immediate solution to this is to update the renv.lock file to use the latest version of `hubValidations` from the GitHub tag. 

The longer-term solution would be to create [an archival `drat` repository](https://cran.r-project.org/web/packages/drat/vignettes/DratStepByStep.html) that contains snapshots from the r-universe.

zkamvar:
I have done three things in this PR that will make the clade generation workflow more stable.

0. pinned the hubverse packages to GitHub releases
1. updated renv and the packages used to the most recent releases
2. pinned the version of R

I updated the renv packages this way:

```r
renv::update()
pkgs <- c("hubverse-org/hubValidations@*release",
  "hubverse-org/hubAdmin@*release",
  "hubverse-org/hubData@*release",
  "hubverse-org/hubUtils@*release"
)
renv::install(pkgs)
renv::snapshot()
```

Note that the hubverse packages for the moment need to be manually
updated in the manner presented. There is a proposal for creating an
archived repository in #137, but until that is created, this will have
to work.

This will fix #137

### DIFF
--- a/.github/workflows/create-modeling-round.yaml
+++ b/.github/workflows/create-modeling-round.yaml
@@ -31,6 +31,7 @@ jobs:
       - name: Set up R 📊
         uses: r-lib/actions/setup-r@v2
         with:
+          r-version: 4.4.1
           install-r: true
           use-public-rspm: true
           extra-repositories: 'https://hubverse-org.r-universe.dev'
--- a/src/renv.lock
+++ b/src/renv.lock
@@ -1,6 +1,6 @@
 {
   "R": {
-    "Version": "4.4.0",
+    "Version": "4.4.1",
     "Repositories": [
       {
         "Name": "CRAN",
@@ -30,7 +30,7 @@
     },
     "MASS": {
       "Package": "MASS",
-      "Version": "7.3-60.2",
+      "Version": "7.3-61",
       "Source": "Repository",
       "Repository": "CRAN",
       "Requirements": [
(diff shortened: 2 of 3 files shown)


bsweger (approved):
Will confess that I don't fully grok the updates to `src/renv/activate.R` script, but also realize that they won't impact the the functionality of the scripts we're running in our variant-nowcast-hub pipelines.

Question: I originally thought that using renv here is a good way to ensure stable, reproducible CI runs. Was that an ill-conceived idea on my part? Is there a point of diminishing returns for using renv? Or is this a single hiccup because we hadn't factoring in the non-archival nature or r-universe?

zkamvar:
> Will confess that I don't fully grok the updates to `src/renv/activate.R` script, but also realize that they won't impact the the functionality of the scripts we're running in our variant-nowcast-hub pipelines.

That's a renv thing. It's the only way it knows how to bootstrap itself. 

> Question: I originally thought that using renv here is a good way to ensure stable, reproducible CI runs. Was that an ill-conceived idea on my part? Is there a point of diminishing returns for using renv? Or is this a single hiccup because we hadn't factoring in the non-archival nature or r-universe?

Single hiccup. The renv entry for the hubValidations package did not include the GitHub metadata, so there was no way to recover hubValidations 0.7.0 after we released 0.8.0 last week.

PR #160, opened 2024-11-15
first draft script to get counts to use for evaluation

elray1:


elray1:
See also https://github.com/reichlab/variant-nowcast-hub/blob/bsweger/sample-target-data-script/src/sample_get_target_data.py -- we should see what ideas we can borrow from there!

src/get_location_date_counts_eval.py
```diff
@@ -0,0 +1,113 @@
...
+        collection_max_date=collection_max_date
+    )
+    
+    logger.info("assign clades")
+    assignments = ct.assign_clades(filtered_metadata)
+    
+    logger.info("get clade counts")
+    counts = assignments.summary.collect()
+    
+    output_file = data_dir / f"{str(reference_date)}_covid_clade_counts.parquet"
```
bsweger, line 105: Double-checking the file format here: if there's any value to providing this time series in human-readable format, it's likely small enough to output as a .csv.

Issue #91, opened 2024-10-03
Data Visualization for Submission Files


trobacker:
We will want some functions to plot forecast (model output) submission files along with observed data. The initial goals are to create ggplot2 functions that be readily applied for quick visualization. This may evolve into interactive tools e.g. Streamlit, Shiny. 

Related to discussion #78.

PR #162, opened 2024-11-15
Basic Visualization Functions

trobacker:
#91 Functions to do some visualization based on a model-output file with sample output type. It plots daily and weekly observed and predicted counts along with logit scale versions using a model output file.


trobacker:
I used 
`plot_summary_graphs(model_output_file = "UMass-HMLR/2024-11-06-UMass-HMLR.parquet", s3_data_date = "2024-11-11")` 
to test an instance locally. There's an option to change the save path, I just used my default to downloads for now.

nickreich:
I made a few changes to the files (see commit notes for details) to try to streamline the code and make it renv compatible.

nickreich:
some late-night notes to self about things to add here:
 - make logit plots compatible when samples aren't present (currently those files are created but without plots, so they can't be opened)
 - make model_output_file argument not dependent on relative file paths that might be different - or standardize expected filepath.
 - add model name to the filename outputs
 - maybe for another PR, but consider adding a plot that has one page per clade and facets by location.

IsaacMacarthur (commented):
Overall it looks really good! I have a few comments though; when I ran the code it outputted a lot of messages that looked like this `geom_smooth()` using method = 'loess' and formula = 'y ~ x' and `summarise()` has grouped output by 'target_date', 'clade'. You can override using the `.groups` argument. which I'm not sure if we want to print, they look a little ugly, but it doesn't affect the actual output. The other suggestion that I would make is if the model submission doesn't submit samples like in the LANL example, don't make the logit pdf's. Right now the function makes empty pdf's if there are no samples, which just get stuck loading if I try to open them. These are both just suggestions and if you think they could be dealt with later or are not important, just let me know and I'll approve the PR.

IsaacMacarthur (approved):

IsaacMacarthur (dismissed):

IsaacMacarthur (approved):

PR #179, opened 2024-11-25
use released version of cladetime

zkamvar:
NOTE: do not merge this until https://github.com/reichlab/cladetime/pull/63 is merged


bsweger (approved):
@zkamvar Just merged https://github.com/reichlab/cladetime/pull/63, so this one can also be merged.

Issue #175, opened 2024-11-20
print fewer messages in viz output


nickreich:
> ...when I ran the code it outputted a lot of messages that looked like this `geom_smooth()` using method = 'loess' and formula = 'y ~ x' and `summarise()` has grouped output by 'target_date', 'clade'. You can override using the `.groups` argument. which I'm not sure if we want to print, they look a little ugly, but it doesn't affect the actual output. 

_Originally posted by @IsaacMacarthur in https://github.com/reichlab/variant-nowcast-hub/pull/162#pullrequestreview-2447118567_

PR #180, opened 2024-11-25
Suppress Visualization Output Messages

trobacker:
#175 Suppress visualization output messages. Doesn't affect actual output. 
Modified a `dplyr` option and added `suppressMessages` wrappers.

### DIFF
--- a/src/plot_summary_graphs.R
+++ b/src/plot_summary_graphs.R
@@ -1,4 +1,3 @@
-options(dplyr.summarise.inform = FALSE) # Suppress message output for dplyr use
 #' save PDF plots: clades by location, daily and weekly, logit and non-logit
 #'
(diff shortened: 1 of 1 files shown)


nickreich (changes requested):
These changes work for me. Is there a reason you added the dplyr options argument at the beginning of the file, rather than inside the function? It worked for me when added inside the function and I think that makes more sense as a place for that option to live. E.g. if we migrate this code to a package at some point, we would need that option to be inside the function itself. I tried moving it to the line immediately following the require(dplyr) in `plot_summary_graphs()` and it seemed to have the same effect.

trobacker:
No good reason I had it there, it should definitely be in the function, thank you! Modified.

Issue #174, opened 2024-11-20
model_summary.R updates


nickreich:
- [ ] roxygenize the function
- [ ] add a line below the "Summary of submissions for..." line that tells which clades are being predicted
- [ ] add an empty line after clades list line

PR #181, opened 2024-11-25
Update Model Summary Output

trobacker:
#174 Added roxygen documentation, and now list clades modeled/predicted.

### DIFF
--- a/src/model_output_summary.R
+++ b/src/model_output_summary.R
@@ -63,13 +63,15 @@ model_summary <- function(target_date, file_path_to_metadata = "../model-metadat
 
     # getting the output types and the model names
     output_types[j] <- paste(unique(model$output_type), collapse = ", ")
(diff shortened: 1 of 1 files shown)


nickreich (changes requested):
Thanks for this update! All the clades are required to be modeled by every team, so my suggestion would be to make it so that:

1. the list of clade names appears just once, below the "Summary of submissions..." top line. this is required to be the same for all teams, so no need to repeat it.
2. sort the list alphabetically (if you wanted to be fanatical about it, put "other" at the end)
3. add a newline to separate these top two lines from the first model summary.

IsaacMacarthur:
Hi, Nick. I just tried to make the changes you requested. Let me know if this is more like what you pictured.

trobacker (commented):
Looks good on my end.

nickreich (approved):
Yes! looks great. thanks.

PR #204, opened 2024-12-10
bsweger/add created at metadata/201

### LINKED ISSUE #201: Add current timestamp to the metadata section of the modeled-clades .json
**Background**

The variant nowcast script creates the weekly `.json` file in `clades_to_model` adds a metadata section to the file to help track the provenence and ensure reproducibility. Example: https://github.com/reichlab/variant-nowcast-hub/blob/main/auxiliary-data/modeled-clades/2024-11-13.json#L12

One item missing from `metadata` is the timestamp (UTC) of when the .json file was created.

**Definition of done**

- [x] The `metadata` section of the weekly `clades_to_model` .json files should contain a top-level `created_at` key with the timestamp of the GitHub workflow run that created it.
- [x] Backfill existing `clades_to_model` with `created_at`, using the timestamps from their corresponding GitHub workflow runs.

bsweger:
Closes #201 

We need to log the timestamp of when the weekly clade list is created (so we can use that date/time to specify the correct reference tree when generating target data).

This PR updates the script that creates the weekly clade .json in `auxiliary-data/modeled-clades`.

It also retroactively adds the new `created_at` field to the clade list for prior rounds (using the timestamps from the associated PRs, which are linked in the individual commit messages).

### DIFF
--- a/auxiliary-data/modeled-clades/2024-10-16.json
+++ b/auxiliary-data/modeled-clades/2024-10-16.json
@@ -8,7 +8,7 @@
         "other"
     ],
     "meta": {
-        "created_at": "2024-10-16T19:37:00+00:00",
+        "created_at": "2024-10-15T19:37:00+00:00",
         "ncov": {
             "schema_version": "v1",
             "nextclade_version": "nextclade 3.8.2",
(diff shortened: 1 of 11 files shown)


auxiliary-data/modeled-clades/2024-10-16.json
```diff
@@ -8,6 +8,7 @@
         "other"
     ],
     "meta": {
+        "created_at": "2024-10-16T19:37:00+00:00",
```
bsweger: This should actually be 10/15, not 10/16 (fixed in a future commit)

PR #209, opened 2024-12-10
Update Model Summary Visualization

trobacker:
Added an argument for LOESS curve visualization, changed LOESS parameters, minor change to visual appearance - (modified `alpha` for points). 

I added a toggle switch to remove the LOESS trend line in the summary plots. The functions for plotting will plot the trend line by default. However, the trend is no longer plotted by default on the daily logit plots when running the main function. 

I experimented with some LOESS parameters with the intention of aiding visual trend analysis. It may turn out they're not good in some situations! 

Lastly, I modified the opaqueness of the points plotted as I believe it's easier to see some details.

### DIFF
--- a/src/plot_summary_graphs.R
+++ b/src/plot_summary_graphs.R
@@ -132,7 +132,8 @@ plot_summary_graphs <- function(
   suppressMessages(save_plots(model_output_data = model_output_logit,
              target_data = target_logit,
              save_path = save_path_daily_logit,
-             page_by_location = page_by_location))
+             page_by_location = page_by_location,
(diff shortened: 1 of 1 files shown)


IsaacMacarthur (commented):
Hi Thomas, this all looks really good! I had just one comment, there doesn't seem to be a way to add back the loess for the daily logits using the plot_summary_graphs function. I'm not sure if we need a way to add back the loess, but it might be worth thinking about.

IsaacMacarthur (approved):
LGTM!

Issue #212, opened 2024-12-11
Create script to generate time series target data for evaluation


bsweger:
**Background**

When we're ready to score the models for a variant-nowcast-hub round, we need produce a time series dataset that assigns SARS-CoV-2 sequences to clades using the correct reference tree and then summarize the results (_i.e._, the [time series version of target data](https://hubverse.io/en/latest/user-guide/target-data.html#time-series))

**Definition of done**

- [x] Script is in repo's `src` directory
- [x] Script contains a function that generates a dataset of clade counts summarized by collection date and location. Columns are location, target_date, clade, observation
- [x] Output contains only clades that were modeled during the nowcast date round
- [x] For locations/target_date/clade combinations with no observations, output a line with observations=0
- [x] Script parameters are: round_id/nowcast_date, sequence_as_of, tree_as_of, min_collection_date, max_collection_date
- [x] `sequence_as_of` defaults to `nowcast_date` + 90 days (but can be overridden)
- [x] `tree_as_of` defaults to `meta.created_at` for the [modeled_clades](https://github.com/reichlab/variant-nowcast-hub/tree/main/auxiliary-data/modeled-clades) json file that corresponds to `nowcast_date`
- [x] time series target data is written to `target-data/time-series/nowcast_date={nowcast_date}/sequence_as_of={sequence_as_of}/part-n.parquet`

Out of scope for this issue is oracle output and automating the script to run each Wednesday. There are separate issues for those tasks:

- #215 
- #216

bsweger:
There's a WIP PR for this here: https://github.com/reichlab/variant-nowcast-hub/pull/160

PR #228, opened 2024-12-20
Create and write oracle output target data

### LINKED ISSUE #215: Create script to generate oracle target data for evaluation
**Background**

When we're ready to score the models for a variant-nowcast-hub round, we need produce target data in [oracle output format](https://hubverse.io/en/latest/user-guide/target-data.html#oracle-output).

The "oracle output" file for each round will be derived from the time series version of the target data. In other words, the time series file needs to be created first, so this issue is dependent on #212 

**Definition of done**

- [ ] Script is in repo's `src` directory
- [ ] Script contains a function that generates an oracle output version of the target data, which will have the following columns: nowcast_date (i.e., round_id), location, target_date, clade, oracle_value. The oracle_value is the count of the clade for the given location and target_date.
- [ ] Script parameter is the location of target time series data that will be used to derive the oracle output and/or the nowcast_date/round_id
- [ ] Oracle target data is saved to `target-data/oracle-output/nowcast_date={nowcast_date}/part-n.parquet`

Out of scope for this issue is automating the script to run each Wednesday (see #216)

bsweger:
Closes #215 

This changeset builds on #212 and uses the time series target data to create oracle output target data that will be used to evaluate models.

The testing notes from the related [time series pull request](https://github.com/reichlab/variant-nowcast-hub/pull/217) apply here as well.

### DIFF
--- a/src/get_target_data.py
+++ b/src/get_target_data.py
@@ -331,7 +331,13 @@ def create_target_data(
         )
     )
 
-    return (time_series_all, pl.LazyFrame())
+    oracle_output = (
+        time_series_all.select(["location", "target_date", "clade", "observation"])
+        .with_columns(pl.lit(nowcast_string).alias("nowcast_date"))
+        .rename({"observation": "oracle_value"})
+    )
+
+    return (time_series_all, oracle_output)
 
 
 def write_target_data(
@@ -341,7 +347,9 @@ def write_target_data(
     # default output directory is the hub's target-data directory
     target_data_dir: Path,
 ) -> tuple[Path, Path]:
-    """Write time series and oracle output target data to S3 bucket."""
+    """Write time series and oracle output target data."""
+
+    # write time series data
     target_time_series_dir = target_data_dir / "time-series"
 
     ts_output_path = (
@@ -357,7 +365,17 @@ def write_target_data(
     time_series.collect().write_parquet(ts_output_path)
     logger.info(f"Target time series saved to {ts_output_path}")
 
-    return (ts_output_path, Path())
+    # write oracle output data
+    target_oracle_output_dir = target_data_dir / "oracle-output"
+    oracle_output_path = target_oracle_output_dir / f"nowcast_date={nowcast_string}"
+    oracle_output_path.mkdir(exist_ok=True, parents=True)
+    oracle_output_path = oracle_output_path / "oracle.parquet"
+
+    oracle = target_data[1]
+    oracle.collect().write_parquet(oracle_output_path)
+    logger.info(f"Target oracle output saved to {oracle_output_path}")
+
+    return (ts_output_path, oracle_output_path)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


src/get_target_data.py
```diff
@@ -357,7 +365,17 @@ def write_target_data(
     time_series.collect().write_parquet(ts_output_path)
     logger.info(f"Target time series saved to {ts_output_path}")
 
-    return (ts_output_path, Path())
+    # write oracle output data
+    target_oracle_output_dir = target_data_dir / "oracle-output"
+    oracle_output_path = target_oracle_output_dir / f"nowcast_date={nowcast_string}"
+    oracle_output_path.mkdir(exist_ok=True, parents=True)
+    oracle_output_path = oracle_output_path / "oracle.parquet"
```
elray1, line 372: I realized that I have a couple of questions about hive partitioning:

1. Is it required to have `n` at the end of the file name, e.g. `"oracle-0.parquet"`?  I'm under the impression that this is necessary, but I wasn't able to find an authoritative source after a few minutes of digging.  If so, we'd also want to change the `timeseries` file name above.
2. If we're partitioning on the `nowcast_date`, do we also need it to be in the data frame itself (as in line 336 above), or is that rendered unnecessary?  I'm not sure what standard/recommended practice is.  It seems like it's helpful to have it in there so that a direct `read_parquet` gets all the columns we might want...

elray1: i guess, looking at this, i wonder if we should also be adding the nowcast date into the timeseries file.

bsweger: @elray1 This is actually a great example of a LDR candidate. When we decide
- on the final filename conversion for oracle and time series target files
- whether or not to add nowcast_date as a column

I'll submit an LDR PR with our reasons (the latter point, especially, was a bit contentious when we debated it for the model-output files on S3)

elray1: Your recommendations for both questions make sense to me.  For the second, I'd want to be sure that if we add the column in, standard hive-aware tools don't run into any issues.  It sounds like you have already checked into that?  If that's all fine, including the column makes sense to me.

I think it would make sense to tie these in to the hubverse discussion about formats/where/how these data are recommended to be stored in hubs.  starting with an LDR for what we've done in the variant nowcast hub sounds good, but/and i think your knowledge about these things should also inform the general hubverse recommendations :)

elray1: so i guess the summary for purposes of this PR is that what's here looks good, but maybe we should go back and also add the nowcast date column into the time series data?

bsweger: Sounds good--do you think we should also add a `sequence_as_of` column to the time series file?

elray1: yeah, that makes sense/would be consistent

elray1 (approved):
i'm approving this pr as is in case you want to merge it in and add the other columns to the time series file in a separate PR.  Also happy for you to make those changes here and I can re-approve it later.

bsweger:
> i'm approving this pr as is in case you want to merge it in and add the other columns to the time series file in a separate PR. Also happy for you to make those changes here and I can re-approve it later.

Thanks--I'll merge it, since I'm trying (with varying degrees of success) to get this script wired up to the GitHub action that automates it. Once I have things working end-to-end, I'll come back and add the columns.

PR #237, opened 2024-12-20
Include partitioning columns in time series target data

bsweger:
This is a follow up to the [conversation on a prior target data PR](https://github.com/reichlab/variant-nowcast-hub/pull/228).

Essentially, we decided to include the fields we're partitioning by in the parquet files for both the oracle and time series target data (the former already included `nowcast_date`, so this PR addresses the time series data only).

### DIFF
--- a/src/get_target_data.py
+++ b/src/get_target_data.py
@@ -323,6 +323,10 @@ def create_target_data(
     time_series_all = (
         all_rows.join(time_series, on=["location", "date", "clade"], how="left")
         .fill_null(strategy="zero")
+        .with_columns(
+            pl.lit(nowcast_string).alias("nowcast_date"),
+            pl.lit(sequence_as_of_string).alias("sequence_as_of"),
+        )
         .rename(
             {
                 "count": "observation",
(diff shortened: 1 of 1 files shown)


src/get_target_data.py
```diff
@@ -485,13 +489,24 @@ def test_target_data():
...
+            "sequence_as_of",
+        ]
+    )
     assert set(ts.columns) == expected_time_series_cols
 
     assert set(ts.get_column("clade").to_list()) == {"AA", "BB", "other"}
     assert ts.get_column("target_date").min() == date(2024, 11, 30)
     assert ts.get_column("target_date").max() == date(2024, 12, 4)
     assert ts.get_column("observation").sum() == 20
+    assert ts.get_column("nowcast_date").unique().to_list() == ["2024-9-11"]
```
bsweger, line 508: @elray1 Do you think it's also worth adding a `tree_as_of` column here, so anyone use this file outside the context of the hub can more easily understand which reference tree was used to get the clade assignments?

elray1: It seems like that could potentially be useful, though not necessarily high priority.  I don't really have strong feelings, but I guess it could be nice to do that.

bsweger: @elray1 ok, stuck `tree_as_of` in there

elray1 (approved):

PR #254, opened 2025-01-07
R Script for Validation Plotting

trobacker:
A prototype R script and sample CladeTime output to plot model output (just UMass-HMLR for now) along with data available at reference date (training) and data validated through CladeTime at a later date - recall rule-of-thumb for validation:  90 days later.

Caution: Don't merge until we've had time to look over this!


src/plot_validation_data.R
```diff
@@ -0,0 +1,122 @@
...
+
+# Meta data for getting data available on reference date
+reference_date <- "2024-10-28" ## REFERENCE DATE
+s3_data_date <- reference_date
+targets_path_s3 <- paste0("https://covid-clade-counts.s3.amazonaws.com/",
+s3_data_date, "_covid_clade_counts.parquet")
+df_retro_oct <- arrow::read_parquet(targets_path_s3)
+
+## load in the hub locations file - convert names/abbreviations
+hub_path = "./"
```
zkamvar, line 24: Declare the packages that need to be loaded and use {here} to indicate where this script lives so we can establish the hub path. 

```suggestion
library("dplyr")
library("ggplot2")
here::i_am("src/plot_validation_data.R")
# Load validation data from CladeTime (pre-made)
# Note: this file is generated through CladeTime as is not made here, must be
# created ahead of time and path changed here
hub_path <- here::here()
df_oct <- read.csv(here::here("auxiliary-data/example-files/summarized_clades_asof_2024-10-28_on_2025-01-07.csv")) |> 
  subset(select = -c(host, country))
# Some locs have missing counts (non-present)
# e.g. subset(df_oct, (date == "2024-10-01") & (location == "TX")) # "MA"
# Resolved with tidyr::complete() 

# Meta data for getting data available on reference date
reference_date <- "2024-10-28" ## REFERENCE DATE
s3_data_date <- reference_date
targets_path_s3 <- paste0("https://covid-clade-counts.s3.amazonaws.com/",
s3_data_date, "_covid_clade_counts.parquet")
df_retro_oct <- arrow::read_parquet(targets_path_s3)

## load in the hub locations file - convert names/abbreviations
```

src/plot_validation_data.R
```diff
@@ -0,0 +1,122 @@
...
+## Modify location names for df_retro
+df_retro_oct <- df_retro_oct %>%
+  left_join(locs, by = "location") %>%
+  # Replace the location names with their abbreviations
+  mutate(location = abbreviation) %>%
+  # Remove the abbreviation column if you no longer need it
+  select(-abbreviation)
+
+# Model output, just UMASS HMLR for now
+df_model_output <- arrow::read_parquet(paste0(hub_path, "model-output/UMass-HMLR/2024-10-30-UMass-HMLR.parquet"))
```
zkamvar, line 39:
```suggestion
df_model_output <- arrow::read_parquet(file.path(hub_path, "model-output/UMass-HMLR/2024-10-30-UMass-HMLR.parquet"))
```

src/plot_validation_data.R
```diff
@@ -0,0 +1,125 @@
...
+  group_by(location, date) |>
+  mutate(total = sum(count)) |>
+  ungroup() |>
+  mutate(value = ifelse(total == 0, 0, count/total)) |>
+  mutate(type = "target")
+
+#######
+
+# Create a PDF file to save the plots
+pdf("~/Downloads/plot_validation_by_location.pdf")
```
IsaacMacarthur, line 60: This line only works if the home directory has a Downloads folder, which isn't a given. it might be good to let the user choose where the pdf is created.

src/plot_validation_data.R
```diff
@@ -0,0 +1,125 @@
...
+# Currently does daily plots, but will be extended to weekly
+library("dplyr")
+library("ggplot2")
+library("arrow")
+here::i_am("src/plot_validation_data.R")
+# Load validation data from CladeTime (pre-made)
+# Note: this file is generated through CladeTime as is not made here, must be
+# created ahead of time and path changed here
+hub_path <- here::here()
+df_validation <- read.csv(here::here("auxiliary-data/example-files/summarized_clades_asof_2024-10-28_on_2025-01-07.csv")) |>
```
zkamvar, line 14: Is this file going to change at all?

src/plot_validation_data.R
```diff
@@ -0,0 +1,125 @@
...
+# created ahead of time and path changed here
+hub_path <- here::here()
+df_validation <- read.csv(here::here("auxiliary-data/example-files/summarized_clades_asof_2024-10-28_on_2025-01-07.csv")) |> 
+  subset(select = -c(host, country))
+# Some locs have missing counts (non-present)
+# e.g. subset(df_validation, (date == "2024-10-01") & (location == "TX")) # "MA"
+# Resolved with tidyr::complete() 
+
+# Meta data for getting data available on reference date
+reference_date <- "2024-10-28" ## REFERENCE DATE
```
zkamvar, line 21: is this reference date always going to be the same?

src/plot_validation_data.R
```diff
@@ -0,0 +1,125 @@
...
+## Modify location names for df_retro
+df_retro <- df_retro %>%
+  left_join(locs, by = "location") %>%
+  # Replace the location names with their abbreviations
+  mutate(location = abbreviation) %>%
+  # Remove the abbreviation column if you no longer need it
+  select(-abbreviation)
+
+# Model output, just UMASS HMLR for now
+df_model_output <- arrow::read_parquet(file.path(hub_path, "model-output/UMass-HMLR/2024-10-30-UMass-HMLR.parquet"))
```
zkamvar, line 42: Is this file name always going to stay the same?

src/plot_validation_data.R
zkamvar: Generally agree. If this script is intended for use by anyone other than @trobacker, it would be good to declare this as a variable at the top and provide instructions for people to change it depending on their machine. 

I've added other comments on static file names that might need to be examined.

src/plot_validation_data.R
```diff
@@ -0,0 +1,125 @@
...
+  ## Check for completion of observations - adds zeros where no clades observed
+  # subset(targets, (date == "2024-10-01") & (location == "MA"))
+  
+  df_out_this_location <- df_model_output |> subset((location == this_location) & (output_type == "mean"))
+  
+  targets_this_location <- targets |> subset(location == this_location)
+  colnames(targets_this_location)[3] <- "clade"
+  
+  p <- ggplot(df_out_this_location, aes(x = target_date, y = value)) + 
+    ggtitle(paste0("Daily Observed and Predicted Proportions \nfor model output in ", this_location, " - 2024-10-30-UMass-HMLR")) +
```
zkamvar, line 92: is this date static?

trobacker: No, it should generally be considered variable and it will be from the model_output files.

src/plot_validation_data.R
trobacker: Not necessarily, in general we will apply it to any model_output file (focusing on UMass HMLR for now).

src/plot_validation_data.R
trobacker: No, it will be changed in the long run. I need to create functions in this script for general use.

src/plot_validation_data.R
trobacker: Yes, it could be any format at this time. They are pre-made files from CladeTime right now to generate validation data.

src/plot_validation_data.R
trobacker: Thanks for the comments! Indeed, this script will have a lot of changes to undergo. I shared it so that folk could see how I plotted training data and validated data through CladeTime for the use case of the UMass-HMLR model. There will absolutely be some work on this file for improvement and general use. Feel free to make commits.

nickreich:
Converted to draft to make changes based on new target data format that will be present shortly.

PR #266, opened 2025-01-10
Explanation of choice to model Nextstrain clades

afmagee42:
This PR attempts to summarize, briefly, the reasons we chose Nextstrain clades as the modeling unit, rather than Pango lineages. 

The topic of aggregating lineages can spiral quickly, so I tried to err on the side of brevity there. If we need more, more can be said.

Also happy for this to be moved around (it has several forward-references, so it may belong later).

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ An algorithmic approach makes the choice of targets more transparent and simplif
 It was with these needs in mind that we chose to use [Nextstrain clades](https://nextstrain.org/blog/2021-01-06-updated-SARS-CoV-2-clade-naming) to define target variants for nowcasting and forecasting.
 Nextstrain clades reflect large-scale trends in SARS-CoV-2 evolution.
 A suitably-sized set of relevant clades to model can be obtained with a simple frequency cutoff.
-[Pango lineages](https://www.nature.com/articles/s41564-020-0770-5) are another approach to naming SARS-CoV-2 variants. By design, they describe both coarse-scale and fine-scale virus evolution. Selecting of a suitable number of relevant lineages for nowcasting and forecasting is not straightforward.
+[Pango lineages](https://www.nature.com/articles/s41564-020-0770-5) are another approach to naming SARS-CoV-2 variants. By design, they describe both coarse-scale and fine-scale virus evolution. Selecting a suitable number of relevant lineages for nowcasting and forecasting is not straightforward.
 That said, as both systems are inherently phylogenetic, there is typically [sufficient](https://raw.githubusercontent.com/nextstrain/ncov-clades-schema/master/clades.svg) correspondence [between them](https://next.nextstrain.org/nextclade/sars-cov-2) that it is possible to model Nextstrain clades but [discuss results in terms of Pango lineages](https://github.com/nextstrain/ncov/blob/master/defaults/clade_display_names.yml).
 For example, Nextstrain clade 24A corresponds to Pango lineage JN.1, 22F to XBB, and 21L to BA.2.[^1]
 
--- a/README.md
+++ b/README.md
@@ -101,7 +101,9 @@ It was with these needs in mind that we chose to use [Nextstrain clades](https:/
 They intentionally follow larger-scale trends in SARS-CoV-2 evolution, and a suitably-sized selection of relevant clades can be obtained with a simple frequency cutoff.
 The somewhat more commonly-encountered [Pango lineages](https://www.nature.com/articles/s41564-020-0770-5) describe variation at a much finer scale, and the selection of a suitable number of relevant variants is non-trivial.
 That said, as both systems are inherently phylogenetic, there is typically [sufficient](https://raw.githubusercontent.com/nextstrain/ncov-clades-schema/master/clades.svg) correspondence [between them](https://next.nextstrain.org/nextclade/sars-cov-2) that it is possible to model Nextstrain clades but [discuss results in terms of Pango lineages](https://github.com/nextstrain/ncov/blob/master/defaults/clade_display_names.yml).
-For example, Nextstrain clade 24A corresponds to Pango lineage JN.1, 22F to XBB, and 21L to BA.2.
+For example, Nextstrain clade 24A corresponds to Pango lineage JN.1, 22F to XBB, and 21L to BA.2.[^1]
+
+[^1]: Note that such correspondence between Nextstrain clades and Pango lineages does not guarantee that a clear and consistent mapping is obvious when looking at the lineages and clades to which individual sequences have been assigned.
 
 ### Prediction horizon
 
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ Further, this system must consistently produce a reasonable number of variants t
 [Nextstrain clades](https://nextstrain.org/blog/2021-01-06-updated-SARS-CoV-2-clade-naming) satisfy both requirements readily.
 They intentionally follow larger-scale trends in SARS-CoV-2 evolution, and a suitably-sized selection of relevant clades can be obtained with a simple frequency cutoff.
 The somewhat more commonly-encountered [Pango lineages](https://www.nature.com/articles/s41564-020-0770-5) describe variation at a much finer scale, and the selection of a suitable number of relevant variants is non-trivial.
-Further, as both systems are inherently phylogenetic, there is [sufficient](https://raw.githubusercontent.com/nextstrain/ncov-clades-schema/master/clades.svg) correspondence [between them](https://next.nextstrain.org/nextclade/sars-cov-2) that it is possible to model Nextstrain clades but [discuss results in terms of Pango lineages](https://github.com/nextstrain/ncov/blob/master/defaults/clade_display_names.yml).
+That said, as both systems are inherently phylogenetic, there is typically [sufficient](https://raw.githubusercontent.com/nextstrain/ncov-clades-schema/master/clades.svg) correspondence [between them](https://next.nextstrain.org/nextclade/sars-cov-2) that it is possible to model Nextstrain clades but [discuss results in terms of Pango lineages](https://github.com/nextstrain/ncov/blob/master/defaults/clade_display_names.yml).
 For example, Nextstrain clade 24A corresponds to Pango lineage JN.1, 22F to XBB, and 21L to BA.2.
 For these reasons, we have chosen to use Nextstrain clades for defining variants.
 
--- a/README.md
+++ b/README.md
@@ -95,14 +95,14 @@ Each week the hub designates up to nine NextStrain clades with the highest repor
 
 #### Why use Nextstrain clades?
 
-To designate a list of modeling targets weekly, it is ideal to have a system that does not require human intervention.
-Further, this system must consistently produce a reasonable number of modeling units to permit the submission of [samples](#probabilistic-forecast-evaluation) of the frequency of every modeling unit in each state for [six weeks](#prediction-horizon).
+To designate, weekly, a list of variants to model, it is ideal to have a system that does not require human intervention.
+Further, this system must consistently produce a reasonable number of variants to permit the submission of [samples](#probabilistic-forecast-evaluation) of the frequency of every variant in each state for [six weeks](#prediction-horizon).
 [Nextstrain clades](https://nextstrain.org/blog/2021-01-06-updated-SARS-CoV-2-clade-naming) satisfy both requirements readily.
 They intentionally follow larger-scale trends in SARS-CoV-2 evolution, and a suitably-sized selection of relevant clades can be obtained with a simple frequency cutoff.
-The somewhat more commonly-encountered [Pango lineages](https://www.nature.com/articles/s41564-020-0770-5) describe variation at a much finer scale, and the selection of a suitable number of relevant modeling units is non-trivial.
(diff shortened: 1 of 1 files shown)


afmagee42:
cc @nickreich, @dylanhmorris

README.md
```diff
@@ -93,6 +93,17 @@ We will not solicit estimates for the US as a whole, in part because evaluating
 
 Each week the hub designates up to nine NextStrain clades with the highest reported prevalence of at least 1% across the US in any of the three complete [USA/CDC epidemiological weeks](https://ndc.services.cdc.gov/wp-content/uploads/MMWR_Week_overview.pdf) (a.k.a. MMWR weeks) preceding the Wednesday submission date. Any clades with prevalence of less than 1% are grouped into an “other” category for which predictions of combined prevalence are also collected. No more than 10 clades (including “other”) are selected in a given week. For details on the workflow that generates this list each week, see the [clade list section](#clade-list) below.
 
+#### Why use Nextstrain clades?
+
+To designate a list of modeling targets weekly, it is ideal to have a system that does not require human intervention.
+Further, this system must consistently produce a reasonable number of modeling units to permit the submission of [samples](#probabilistic-forecast-evaluation) of the frequency of every modeling unit in each state for [six weeks](#prediction-horizon).
```
nickreich, line 99: I think you are using "modeling unit" to describe the variant/lineage/clade thing, but I find it hard to parse. (And in the first sentence the term "modeling target" is used.) What if we replaced "modeling unit" here, and below on line 102, with "variant". We could offer a mea culpa for that being an imprecise biological term if it makes us feel less squeamish about it.

README.md
```diff
@@ -93,6 +93,17 @@ We will not solicit estimates for the US as a whole, in part because evaluating
...
+[Nextstrain clades](https://nextstrain.org/blog/2021-01-06-updated-SARS-CoV-2-clade-naming) satisfy both requirements readily.
+They intentionally follow larger-scale trends in SARS-CoV-2 evolution, and a suitably-sized selection of relevant clades can be obtained with a simple frequency cutoff.
+The somewhat more commonly-encountered [Pango lineages](https://www.nature.com/articles/s41564-020-0770-5) describe variation at a much finer scale, and the selection of a suitable number of relevant modeling units is non-trivial.
+Further, as both systems are inherently phylogenetic, there is [sufficient](https://raw.githubusercontent.com/nextstrain/ncov-clades-schema/master/clades.svg) correspondence [between them](https://next.nextstrain.org/nextclade/sars-cov-2) that it is possible to model Nextstrain clades but discuss results in terms of Pango lineages.
```
nickreich, line 103: Maybe worth adding on parenthetically at the end of this sentence something like ", even if there is not always a perfect one-to-one alignment between the clade and lineage assignments for a single sequence."

README.md
```diff
@@ -93,6 +93,17 @@ We will not solicit estimates for the US as a whole, in part because evaluating
...
+For example, Nextstrain clade 24A corresponds to Pango lineage JN.1, 22F to XBB, and 21L to BA.2.
```
nickreich, line 104: We could explicitly point to [this file](https://github.com/nextstrain/ncov/blob/master/defaults/clade_display_names.yml)?

nickreich (changes requested):
Thanks for putting this together! A few minor suggestions included here.

README.md
afmagee42: I am fine with "variant" if you think it's clearer.

README.md
afmagee42: I like this in principle, but I'm having some trouble seeing how to be handwavy enough here not to open up a(t least one) rather large can of worms requiring potentially a fair bit more text to explain.

In a perfect world there would be a many-to-one mapping such that every Pango lineage corresponds to exactly one Nextstrain clade (not a one-to-one mapping). But that quickly leads to the cans of worms that are nestedness of naming (which is why the perfect world is many-to-one) and "lineage assignments aren't data" (which is why we don't live in a perfect world), and if we're not careful the pain that is the [ARG](https://pmc.ncbi.nlm.nih.gov/articles/PMC10796009/).

afmagee42: Perhaps something like one of these? (Which add load-bearing phrases to accompany the heavy lifting done by "sufficient correspondence.")

1. Further, as both systems are inherently phylogenetic, there is [sufficient](https://raw.githubusercontent.com/nextstrain/ncov-clades-schema/master/clades.svg) correspondence [between them](https://next.nextstrain.org/nextclade/sars-cov-2) that it is broadly possible to model Nextstrain clades but discuss results in terms of Pango lineages (even if, at the level of the assignment of individual sequences, the correspondence appears somewhat murkier).
2. Further, as both systems are inherently phylogenetic, there is [sufficient](https://raw.githubusercontent.com/nextstrain/ncov-clades-schema/master/clades.svg) correspondence [between them](https://next.nextstrain.org/nextclade/sars-cov-2) that it is broadly possible to model Nextstrain clades but discuss results in terms of Pango lineages (even if the mappings are not perfectly clean when looking at sequence-level assignments of clades and lineages).

README.md
```diff
@@ -93,6 +93,17 @@ We will not solicit estimates for the US as a whole, in part because evaluating
...
 
+#### Why use Nextstrain clades?
+
+To designate, weekly, a list of variants to model, it is ideal to have a system that does not require human intervention.
+Further, this system must consistently produce a reasonable number of variants to permit the submission of [samples](#probabilistic-forecast-evaluation) of the frequency of every variant in each state for [six weeks](#prediction-horizon).
+[Nextstrain clades](https://nextstrain.org/blog/2021-01-06-updated-SARS-CoV-2-clade-naming) satisfy both requirements readily.
+They intentionally follow larger-scale trends in SARS-CoV-2 evolution, and a suitably-sized selection of relevant clades can be obtained with a simple frequency cutoff.
+The somewhat more commonly-encountered [Pango lineages](https://www.nature.com/articles/s41564-020-0770-5) describe variation at a much finer scale, and the selection of a suitable number of relevant variants is non-trivial.
```
dylanhmorris, line 102: Perhaps give an example here?  
```suggestion
[Pango lineages](https://www.nature.com/articles/s41564-020-0770-5) are another approach to naming SARS-CoV-2 variants. By design, they describe both coarse-scale and fine-scale virus evolution. Selecting an appropriate set of non-overlapping Pango lineages is less straightforward, as it requires first determining an appropriate scale of variation for modeling.
```

README.md
```diff
@@ -93,6 +93,17 @@ We will not solicit estimates for the US as a whole, in part because evaluating
...
+Further, as both systems are inherently phylogenetic, there is [sufficient](https://raw.githubusercontent.com/nextstrain/ncov-clades-schema/master/clades.svg) correspondence [between them](https://next.nextstrain.org/nextclade/sars-cov-2) that it is possible to model Nextstrain clades but [discuss results in terms of Pango lineages](https://github.com/nextstrain/ncov/blob/master/defaults/clade_display_names.yml).
```
dylanhmorris, line 103:
```suggestion
That said, as both systems are inherently phylogenetic, there is typically [sufficient](https://raw.githubusercontent.com/nextstrain/ncov-clades-schema/master/clades.svg) correspondence [between them](https://next.nextstrain.org/nextclade/sars-cov-2) that it is possible to model Nextstrain clades but [discuss results in terms of Pango lineages](https://github.com/nextstrain/ncov/blob/master/defaults/clade_display_names.yml).
```

README.md
```diff
@@ -93,6 +93,17 @@ We will not solicit estimates for the US as a whole, in part because evaluating
...
+For example, Nextstrain clade 24A corresponds to Pango lineage JN.1, 22F to XBB, and 21L to BA.2.
+For these reasons, we have chosen to use Nextstrain clades for defining variants.
```
dylanhmorris, line 105:
```suggestion
For these reasons, we have chosen to use Nextstrain clades as our target units for nowcasting and forecasting.
```

dylanhmorris (commented):
Thanks @afmagee42. A few questions and a few wording suggestions.

README.md
dylanhmorris: I suggested just adding "typically possible", which I think captures it. The point about single-sequence-level classification is helpful context, but I'd suggest relegating it to a footnote.

README.md
```diff
@@ -93,6 +93,17 @@ We will not solicit estimates for the US as a whole, in part because evaluating
...
 
+#### Why use Nextstrain clades?
+
+To designate, weekly, a list of variants to model, it is ideal to have a system that does not require human intervention.
+Further, this system must consistently produce a reasonable number of variants to permit the submission of [samples](#probabilistic-forecast-evaluation) of the frequency of every variant in each state for [six weeks](#prediction-horizon).
+[Nextstrain clades](https://nextstrain.org/blog/2021-01-06-updated-SARS-CoV-2-clade-naming) satisfy both requirements readily.
+They intentionally follow larger-scale trends in SARS-CoV-2 evolution, and a suitably-sized selection of relevant clades can be obtained with a simple frequency cutoff.
```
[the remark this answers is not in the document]
afmagee42: I think "relevant" is doing some important work here. It's very easy to define a bad set of Pango lineages for modeling, and not straightforward to define a useful set. But even with NextStrain clades, I could pretty easily define a _bad_ set of clades (plus other), say, take the first-named 9 (19 A-B, 20A-H), producing nearly 100% other.

It is incredibly convenient that NextStrain clades allow us to only ever think about the most-specific names available to any sequence, and then take the most frequent ones, and model those.

README.md
afmagee42: I think we can just leave this off, with the changed sentence above, and end with the "they're comparable enough" sentences.

README.md
```diff
@@ -93,6 +93,17 @@ We will not solicit estimates for the US as a whole, in part because evaluating
 
 Each week the hub designates up to nine NextStrain clades with the highest reported prevalence of at least 1% across the US in any of the three complete [USA/CDC epidemiological weeks](https://ndc.services.cdc.gov/wp-content/uploads/MMWR_Week_overview.pdf) (a.k.a. MMWR weeks) preceding the Wednesday submission date. Any clades with prevalence of less than 1% are grouped into an “other” category for which predictions of combined prevalence are also collected. No more than 10 clades (including “other”) are selected in a given week. For details on the workflow that generates this list each week, see the [clade list section](#clade-list) below.
 
+#### Why use Nextstrain clades?
+
+To designate, weekly, a list of variants to model, it is ideal to have a system that does not require human intervention.
+Further, this system must consistently produce a reasonable number of variants to permit the submission of [samples](#probabilistic-forecast-evaluation) of the frequency of every variant in each state for [six weeks](#prediction-horizon).
```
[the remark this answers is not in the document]
afmagee42: Fair enough, re, ordering, and I do like the explanation of why algorithmic is good.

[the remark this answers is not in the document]
afmagee42: Maybe, accounting for this and the size bit, we get this?
```suggestion
The Hub must define a list of variants to model each week.
This system should consistently produce a reasonable number of distinct variants as modeling targets.
(Experimentation showed this to be approximately 10 variants or fewer in order to accommodate a sufficient number of [samples](#probabilistic-forecast-evaluation) of the frequency of every variant in each state for [six weeks](#prediction-horizon).)
Ideally, it should be algorithmic and operate without human intervention.
An algorithmic approach makes the choice of targets more transparent and simplifies Hub administration.
```

README.md
afmagee42: I know I'm doing a lot of hand-waving here, but as mentioned above, we don't want to require an entire explainer of lineage aggregation just to make this point.

To that end, I'm also:
- Iffy on diving into the lower-and-higher-order nature of naming schemes. Nested names are messy, but both Nextstrain clades and Pango lineages do it.
- Not sold on "non-overlapping" because when you read lists of aggregated Pango lineages, or our list of clades to model, implicit paraphyly is pretty much guaranteed (and it is _implicit_).

README.md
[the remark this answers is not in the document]
dylanhmorris: Sounds good.

README.md
afmagee42: I like "typically possible" and I like that a footnote affords a bit more space for clarity in this aside. I have done so.

README.md
nickreich: I don't have strong feelings about the language here. I'd be ok keeping the existing text (not taking @dylanhmorris 's suggestion), because it feels shorter/simpler.

README.md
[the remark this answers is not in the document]
nickreich: I like the suggested text from @afmagee42 .

nickreich (commented):
I think this looks good, and suggest resolving the two remaining suggestions and then having it be "good to go".

README.md
dylanhmorris: Alternative suggestion that stays closer to the existing text but avoids what I think are some potential pitfalls.
```suggestion
[Pango lineages](https://www.nature.com/articles/s41564-020-0770-5) are another approach to naming SARS-CoV-2 variants. By design, they describe both coarse-scale and fine-scale virus evolution. Selecting of a suitable number of relevant variants for nowcasting and forecasting is not straightforward.
```

dylanhmorris: Rationale: I'd like to avoid "non-trivial" and to avoid giving the impression that Pango doesn't do coarse-scale.

README.md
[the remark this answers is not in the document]
dylanhmorris: Me too. I think let's go with that!

dylanhmorris:
Agree with @nickreich. @afmagee42 feel free to take or leave my final suggestion on Pango.

README.md
afmagee42: This is a good compromise.

afmagee42:
Added suggestions, made one little parallelism cleanup and I think we're set

nickreich (approved):

nlinton:
Appreciate you putting this together!

PR #267, opened 2025-01-10
Bs/fix target date schema/265

### LINKED ISSUE #265: Force Polars to type text columns as "string" instead of "large string" when writing target data files

bsweger:
Resolves #265 

**Background**

We want Arrow to read `nowcast_date` and `sequence_as_of` colummns as a `string` data type rather than `large_string` (see issue linked above for more details).

This PR changes the method of writing both oracle and time series target data parquet files to ensure that all text columns will be read by arrow as `string`.

**Testing**

The tests in `get_target_data.py` now read the target data files with PyArrow to confirm the schema.

Some manual poking around confirmed this in R.

Reading a single time series file before the change (partitioned reads didn't work, which is the reason for the PR)

```
FileSystemDataset with 1 Parquet file
7 columns
target_date: date32[day]
location: large_string
clade: large_string
observation: uint32
nowcast_date: large_string
sequence_as_of: large_string
tree_as_of: large_string
```

Reading partitioned time series files after the change:

```
> ds <- arrow::open_dataset("/Users/rsweger/code/variant-nowcast-hub/target-data/time-series/", format="parquet")
> ds
FileSystemDataset with 2 Parquet files
7 columns
target_date: date32[day]
location: string
clade: string
observation: int32
nowcast_date: string
sequence_as_of: string
tree_as_of: string
```

### DIFF
--- a/src/get_target_data.py
+++ b/src/get_target_data.py
@@ -629,17 +629,19 @@ def test_target_data_integration(caplog, tmp_path):
     assert len(modeled_clades) == len(ts_clades)
     assert set(ts_clades) == (set(modeled_clades))
 
-    assert ts.get_column("tree_as_of").unique().to_list() == ["2024-09-09"]
+    assert ts.get_column("tree_as_of").unique().to_list() == [
+        datetime(2024, 9, 9).date()
+    ]
 
     # check time series column data types
     ts_schema_dict = ts.schema.to_python()
     assert ts_schema_dict.get("location") is str
     assert ts_schema_dict.get("target_date") is date
     assert ts_schema_dict.get("clade") is str
-    assert ts_schema_dict.get("observation") is int
-    assert ts_schema_dict.get("nowcast_date") is str
-    assert ts_schema_dict.get("sequence_as_of") is str
-    assert ts_schema_dict.get("tree_as_of") is str
+    assert ts_schema_dict.get("observation") is float
+    assert ts_schema_dict.get("nowcast_date") is date
+    assert ts_schema_dict.get("sequence_as_of") is date
+    assert ts_schema_dict.get("tree_as_of") is date
 
     # time series rows should = total target dates * total locations * total clades
     len(target_dates) * len(state_list) * len(modeled_clades) == ts.height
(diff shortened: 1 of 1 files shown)


trobacker (approved):

src/get_target_data.py
```diff
@@ -367,10 +380,22 @@ def write_target_data(
...
+
+    ts_schema = pa.schema(
+        [
+            ("target_date", pa.date32()),
+            ("location", pa.string()),
+            ("clade", pa.string()),
+            ("observation", pa.int32()),
+            ("nowcast_date", pa.string()),
+            ("sequence_as_of", pa.string()),
+            ("tree_as_of", pa.string()),
```
elray1, line 394: should these be a `pa.date32()` data type?

src/get_target_data.py
```diff
@@ -380,7 +405,19 @@ def write_target_data(
...
-    oracle.collect().write_parquet(oracle_output_path)
+    oracle_arrow = oracle.collect().to_arrow()
+
+    oracle_schema = pa.schema(
+        [
+            ("location", pa.string()),
+            ("target_date", pa.date32()),
+            ("clade", pa.string()),
+            ("oracle_value", pa.int32()),
+            ("nowcast_date", pa.string()),
```
elray1, line 416: should this be a `pa.date32()` data type?

src/get_target_data.py
elray1: This may be addressed with an answer of "no" in your comment on the unit tests below, because it's being used as a partition key?  If so, maybe worth adding a comment here explaining the reason for string data type for these date columns.

elray1 (changes requested):
Looks good, though I had one question about using `string` as the data type for some columns that are dates.  Requesting a change to either update the data types to dates, or add a comment where the schema is defined explaining why we can't use dates.

src/get_target_data.py
bsweger: You're right that the dates used as partition keys should be strings to prevent a data type mismatch when being read by Arrow's `open_dataset()` function. Additionally, `nowcast_date` is a string in `tasks.json` (I think? The values are in quotes), and the target data format RFC states that column types should match their config counterparts.

Which makes me wonder if `target_date` should also be string?

So we're left with `sequence_as_of` and `tree_as_of`. Those have been strings in the target data PRs we've been merging all along, though it wasn't explicit. These two columns aren't part of the hub config, so we can set their types to whatever is convenient. Would it be helpful for downstream analysis if they were `pa.date32`?

src/get_target_data.py
bsweger: See above comment with a follow-up question about the data format here.

src/get_target_data.py
elray1: it seems so unfortunate that partitioning on a particular column impacts what data types you're allowed to use for that column!

elray1: i'm now inclined to just say all of these date columns should be stored as a string?? It seems better to be consistent about this than to have different data types for different columns, right?

elray1: I confirmed that when directly reading in the model output files with polars (skipping arrow), the column data types are what you listed above.  I think that it would be really ideal if we could figure out how to get it so that at least those 7 columns (other than model_id) to have the same data type in the target/oracle output data and the model output data.

> When not provided with an explicit schema, Arrow will apply a string data type to the partitioning column. If that column also exists in the parquet file itself, the data type must match.

I would expect that any arrow-based tools we eventually build for loading would eventually apply a schema for the oracle output (this is trickier for the nowcast data). Does that resolve this problem, so that we could save the things with a date data type and expect anyone using arrow to load them in to specify a schema with date data types?

bsweger: > I would expect that any arrow-based tools we eventually build for loading target data would apply a schema for the oracle output (this is trickier for the nowcast data). Does that resolve this problem, so that we could save the things with a date data type and expect anyone using arrow to load them in to specify a schema with date data types?

Yes! If we're operating under the assumption that Hubverse tools will a apply a schema when reading target data, then this script can type everything as you suggest (i.e., match the model-output schema).

Just pushed some changes for this.

elray1 (approved):
lgtm!  thanks for your patience with my severe waffling about how to best set all this up

PR #289, opened 2025-01-24
Energy Score Functions for Model Evaluation

trobacker:
R script containing functions to get energy scores from model output and produce basic summaries of scores.

### DIFF
--- a/src/model_scoring_functions.R
+++ b/src/model_scoring_functions.R
@@ -23,8 +23,8 @@ get_energy_scores <- function(
                           model_output_file = model_output_file,
                           ref_date = as.Date(ref_date))
   
-  df_scores <- calc_energy_scores(targets = as.data.frame(data[1]), 
-                                  df_model_output = as.data.frame(data[2]))
+  df_scores <- calc_energy_scores(targets = data[[1]], 
+                                  df_model_output = data[[2]])
   
   return(df_scores)
 }
--- a/src/model_scoring_functions.R
+++ b/src/model_scoring_functions.R
@@ -20,14 +20,14 @@ get_energy_scores <- function(
   require("readr")
   require("arrow")
   require("scoringRules")
-
+  
   data <- process_target_data(hub_path = hub_path,
-                          model_output_file = model_output_file,
-                          ref_date = as.Date(ref_date))
-
+                              model_output_file = model_output_file,
+                              ref_date = as.Date(ref_date))
+  
   df_scores <- calc_energy_scores(targets = data[[1]],
                                   df_model_output = data[[2]])
-
+  
   return(df_scores)
 }
 
@@ -46,58 +46,58 @@ process_target_data <- function(hub_path = here::here(),
   df_model_output <- arrow::read_parquet(file.path(hub_path, "model-output", model_output_file))
   clades <- sort(unique(df_model_output$clade))
   locs_modeled <- sort(unique(df_model_output$location))
-
+  
   # Load validation data
   oracle_file <- paste0("nowcast_date=", ref_date)
   oracle_path <- file.path(hub_path, "target-data", "oracle-output",
                            oracle_file, "oracle.parquet")
   df_validation <- arrow::read_parquet(oracle_path)
-
+  
   # Pick out dates to score that were not used for any training
   unscored_file <- paste0(ref_date, ".csv")
   unscored_path <- file.path(hub_path, "auxiliary-data", "unscored-location-dates", unscored_file)
   df_unscored <- read_csv(unscored_path,
                           show_col_types = FALSE)
-
+  
   # Load location data to match abbreviations to full name locations
   load(file.path(hub_path, "auxiliary-data", "hub_locations.rda"))
   locs_join <- hub_locations |>
     dplyr::select(abbreviation, location_name) |>
     rename(location = location_name)
-
+  
   df_unscored <- df_unscored |>
     left_join(locs_join, by = c("location" = "location")) |>
     select(abbreviation, target_date, count) |>
     rename(location = abbreviation)
-
+  
   # Look at subset where count == 0
   df_unscored_zeros <- subset(df_unscored , count == 0)
-
+  
   targets <- df_validation
-
+  
   # Max forecast date
   forecast_date <- ref_date + 10
-
+  
   # Keep all forecast data
   targets_forecast_temp <- targets |>
     subset(target_date > ref_date & target_date <= forecast_date)
-
+  
   # Nowcast temp data frame to be refined
   targets_nowcast_temp <- targets |>
     subset(target_date <= ref_date)
-
+  
   # Join by target_date and location - keeping only those
   # Merging "unscored" location data with full target data
   # Need only be done for dates <= ref_date
   targets_nowcast_temp <- targets_nowcast_temp |>
     right_join(y = df_unscored_zeros, by = join_by(target_date, location)) |>
     select(-count)
-
+  
   # Combine refined nowcast data with forecast data
   targets <- rbind(targets_nowcast_temp, targets_forecast_temp) |>
     subset(location %in% locs_modeled) |>
     arrange(location, target_date)
-
+  
   return(list(targets, df_model_output))
 }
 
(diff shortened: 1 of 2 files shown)


src/model_scoring_functions.R
```diff
@@ -0,0 +1,219 @@
...
+  require("readr")
+  require("arrow")
+  require("scoringRules")
+  
+  data <- process_target_data(hub_path = hub_path,
+                          model_output_file = model_output_file,
+                          ref_date = as.Date(ref_date))
+  
+  df_scores <- calc_energy_scores(targets = as.data.frame(data[1]), 
+                                  df_model_output = as.data.frame(data[2]))
```
zkamvar, line 27: Minor note:

If the output of `process_target_data` is a list of two data frames, then there's no reason to convert the lists to data frames:

```suggestion
  df_scores <- calc_energy_scores(targets = data[[1]], 
                                  df_model_output = data[[2]])
```

src/model_scoring_functions.R
```diff
@@ -0,0 +1,219 @@
...
+get_energy_scores <- function(
+    hub_path = "../",
+    model_output_file = NULL,
+    ref_date = NULL
+){
+  require("tidyr")
+  require("dplyr")
+  require("readr")
+  require("arrow")
+  require("scoringRules")
```
zkamvar, line 20: A minor note, but may be more work than its worth:

loading packages in a function is not good practice and may cause headaches down the line (i.e. dplyr masks some functions from stats and base).

I'd be happy to help rectify this.

src/model_scoring_functions.R
trobacker: Yeah, I thought so too! But on my local device, it made `targets` and `df_model_output` have `NULL` values for some reason. I'll see if it was a fluke and try it again. Thanks!

trobacker: Maybe a `[[ ]]`?

trobacker: Ah, I see now. A double bracket `[[ ]]` will resolve that. Thanks, Zhian!

src/model_scoring_functions.R
```diff
@@ -0,0 +1,219 @@
...
+  
+  locs <- sort(unique(targets$location))
+  dates <- sort(unique(targets$target_date))
+  
+  # Loop over each day at each location
+  for(loc in locs){
+    for(day in dates){
+      
+      # For debugging and satisfaction
+      if(display = TRUE){
```
zkamvar, line 123: This causes the script to fail

```suggestion
      if(display == TRUE){
```

zkamvar: actually, this whole clause should go because `display` is not a parameter

trobacker: Indeed, it can go. A left over relic.

rogersbw (changes requested):
I believe the scoring procedure used here is slightly different from what is discussed here:

https://github.com/reichlab/variant-nowcast-hub?tab=readme-ov-file#background

Happy to meet and discuss on Monday! Overall great coding though!

trobacker:
Hi @rogersbw, I modified the ES calculation to include 10000 samples, 100 (posterior draws) * 100 samples of multinomial counts, on each day/date. Do you mind reviewing that piece (last commit) and confirming that's correct? 

It certainly takes longer to run now and I'll plan on optimizing the code next - i.e. replacing for loops with say functional programming techniques. I want to be sure the ES algorithm is correct first. Thanks!

src/model_scoring_functions.R
```diff
@@ -0,0 +1,219 @@
...
+      samp_multinomial_counts <- matrix(nrow = dim(samp_matrix)[1], ncol = 0)
+      
+      for(col in 1:dim(samp_matrix)[2]){
+        
+        # Get sample clade proportions from predictive distribution
+        samp_props <- samp_matrix[,col]
+        
+        ## Generate 100 multinomial observations for samp_props
+        # Need the N for each loc/day from the validation data
+        N <- sum(subset(targets,
```
rogersbw, line 171: This can be moved outside of the loop, as for each combination of loc and day, the N will stay constant through this loop.

rogersbw: Also, how are day/loc combinations with 0 observations handled? As I'm running through this, it seems to be keeping in the 0 observation days and giving them an energy score of 0 (perfect score). I believe we will want these to be NA values instead.

rogersbw (commented):
I have one question about handling location date combinations with 0 observed counts. See the comment within the code.

src/model_scoring_functions.R
trobacker: Thanks for the comments, Ben! I modified the energy scores to have `NA` when the observed counts sum to 0. I also moved the `N` assignment. I focus on code optimization next.

src/model_scoring_functions.R
```diff
@@ -0,0 +1,223 @@
...
+    rename(location = location_name)
+  
+  df_unscored <- df_unscored |>
+    left_join(locs_join, by = c("location" = "location")) |>
+    select(abbreviation, target_date, count) |>
+    rename(location = abbreviation)
+  
+  # Look at subset where count == 0
+  df_unscored_zeros <- subset(df_unscored , count == 0)
+
```
rogersbw, line 74:
```suggestion
```

src/model_scoring_functions.R
```diff
@@ -0,0 +1,223 @@
...
+  forecast_date <- ref_date + 10
+  
+  # Keep all forecast data
+  targets_forecast_temp <- targets |>
+    subset(target_date > ref_date & target_date <= forecast_date)
+  
+  # Nowcast temp data frame to be refined
+  targets_nowcast_temp <- targets |>
+    subset(target_date <= ref_date)
+
```
rogersbw, line 87: Do we need to separate out by forecasts and nowcasts?

rogersbw: Also, I believe that the oracle data should just have the dates that need to be scored? So we don't necessarily need to filter the dates, although it's not a bad idea to make sure all dates are within ref_date - 31 and ref_date + 10.

rogersbw, line 87:
```suggestion
```

src/model_scoring_functions.R
```diff
@@ -0,0 +1,223 @@
...
+  # Merging "unscored" location data with full target data
+  # Need only be done for dates <= ref_date
+  targets_nowcast_temp <- targets_nowcast_temp |>
+    right_join(y = df_unscored_zeros, by = join_by(target_date, location)) |>
+    select(-count)
+  
+  # Combine refined nowcast data with forecast data
+  targets <- rbind(targets_nowcast_temp, targets_forecast_temp) |>
+    subset(location %in% locs_modeled) |>
+    arrange(location, target_date)
```
rogersbw, line 98:
```suggestion
  # Remove location date combinations from unscored-location-dates file and subset to only locations modeled
  targets <- df_validation |>
    anti_join(y = df_unscored, by = join_by(target_date, location)) |>
    subset(location %in% locs_modeled) |>
    arrange(location, target_date)
```

rogersbw: I think this should do what we're looking for?

rogersbw (changes requested):
Sorry to keep sending it back! I had not well thought about handling the unscored location/dates until I saw your message on slack. I believe these corrections should do what we want?

src/model_scoring_functions.R
rogersbw: Do we need to separate out by forecasts and nowcasts?

rogersbw: Also, I believe that the oracle data should just have the dates that need to be scored? So we don't necessarily need to filter the dates, although it's not a bad idea to make sure all dates are within ref_date - 31 and ref_date + 10.

trobacker:
As discusses in our modeling meeting today, I'm going to have the energy scores available for all dates/locs and build another function that will filter for ones we've generally agreed to use for final model scoring according to the hub evaluation scheme. 
The motivation for this is that that some modelers express interest in getting scores for all data used though we will exclude some locs/dates for final scoring on the hub.

trobacker:
FWIW, I wrote a different version that I would have expected to be computationally faster which uses `expand.grid` to avoid nested for-loops, and an `apply` to sample the multinomials faster, but it actually took 5 seconds longer than my latest commit.

rogersbw:
Looks good, merge away!

rogersbw (approved):

PR #304, opened 2025-01-30
January 29 nowcast

Ehsan-suez:


IsaacMacarthur (approved):

nickreich:
@Ehsan-suez In the future, please try to submit PRs by 8pm ET on Wednesdays. We start to run jobs then (like rebuilding the dashboard and generating target data files), and if late submissions come in it means that we have to re-run some of these actions. Thanks as always for your submissions to the hub!

PR #315, opened 2025-02-10
[chore] address security issues flagged by GitHub CodeQL

### LINKED ISSUE #312: Fix security issues flagged by GitHub CodeQL
**Background**

We recently enabled security scanning on the variant-nowcast-hub repo, and the initial scan flagged some things to fix: https://github.com/reichlab/variant-nowcast-hub/security/code-scanning

There are 17 alerts, but only two root causes, both of which are straightforward to fix.

**Definition of done**

- [ ] All workflows have explicit permissions
- [ ] All workflows reference third-party GitHub actions by commit hash instead of by tag

bsweger:
Closes #312 

**Background**

See the ticket for a summary of two issue types addressed in this PR.

**Testing**

- This PR can be reviewed commit by commit (there are two commits: one for each issue type)
- I tested the four actions with updated permissions in a [fork of variant-nowcast-hub](https://github.com/bsweger/variant-nowcast-hub/actions)

### DIFF
--- a/.github/workflows/cache-hubval-deps.yaml
+++ b/.github/workflows/cache-hubval-deps.yaml
@@ -1,4 +1,7 @@
 name: Build dependency cache on main branch
+permissions:
+  contents: read
+
 on:
   push:
     branches: [main, master]
--- a/.github/workflows/cache-hubval-deps.yaml
+++ b/.github/workflows/cache-hubval-deps.yaml
@@ -17,7 +17,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: r-lib/actions/setup-r@v2
+      - uses: r-lib/actions/setup-r@473c68190595b311a74f208fba61a8d8c0d4c247  #v2.11.1
         with:
           install-r: true
           use-public-rspm: true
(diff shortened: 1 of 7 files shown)


zkamvar (dismissed):
LGTM.

bsweger:
@zkamvar Thanks for taking a look and for the more nuanced conversation about GitHub action security in Slack.

I moved this PR to WIP status so we can pause while I try to assimilate the bigger picture into an RFC.

bsweger:
@zkamvar I decided to get this merged and let it bake for a bit...especially to see how dependabot handles the actions that are pinned by commit hash. I _think_ all of our third party actions use semantic versioning, but I'm not quite sure about the r-lib actions.

zkamvar (approved):
LGTM

> I think all of our third party actions use semantic versioning, but I'm not quite sure about the r-lib actions.

r-lib actions definitely use semantic versioning. 

> I decided to get this merged and let it bake for a bit...especially to see how dependabot handles the actions that are pinned by commit hash. 

I really wish their documentation were more clear on this point `-_-`

I mean, if dependabot doesn't handle the ones that are pinned by hash, we won't find out. I think the only way to find out is to make a a couple of throwaway repos that contains action pinned to version that has a known bug. One repo uses the semver tag and one uses the hash. If we get an alert for both, then it's safe to use the hash.

PR #333, opened 2025-02-26
[AUTO] Adding 2025-02-26-LANL-CovTransformer.parquet with predictions

MKupperman:
> The views expressed here are those of the authors and does not necessarily represent the official views of the National Institutes of Health, Los Alamos National Laboratory, or the US Government.


MKupperman:
@IsaacMacarthur - do you understand this error message from the validator, 

>  Required task ID/output type/output type ID combinations missing.

Here's the first few rows:

```

location | clade | value | target_date | output_type | output_type_id | nowcast_date
-- | -- | -- | -- | -- | -- | --
AZ | recombinant | 0.0 | 2025-01-26 | mean | <NA> | 2025-02-26
AZ | recombinant | 0.0 | 2025-01-27 | mean | <NA> | 2025-02-26
AZ | recombinant | 0.0 | 2025-01-28 | mean | <NA> | 2025-02-26
AZ | recombinant | 0.0 | 2025-01-29 | mean | <NA> | 2025-02-26
AZ | recombinant | 0.0 | 2025-01-30 | mean | <NA> | 2025-02-26
```

My code for this hasn't changed since last week (and that submission was accepted), and I don't see any changes in the last week to this hub that would impact this. Maybe something upstream?

trobacker:
Hi there, so I was able to reproduce your error locally. There haven't been any updates to `hubValidations` since January 22 and no changes on the Hub that would affect the submission validation checks. 

Short answer, I don't know why that error is happening right now. Validation does pass for the UMass submission and the only difference I can tell now is that the UMass submission includes samples and mean output types while LANL is mean only. It seems like it's something in the submission file and not the Hub validation checks. 

I'll keep investigating and ask other folk.

trobacker:
Success!

trobacker (approved):

MKupperman:
There we go! Issue was in 24D clade. Our model wasn't making predictions (need to debug that later) for that lineage. 

Maybe not the error message to expect, but that seems to have been the cause.

PR #343, opened 2025-03-05
Update target-data files

### LINKED ISSUE #296: Update variant nowcast target data to reflect the latest standards and decicions
## Background

In the weeks since we began generating target data for the variant nowcast hub, we've identified two changes to make.

### oracle output

Because the oracle output files are not partitioned by `sequence_as_of`, they will be overwritten each time the hub's [post-submission jobs](https://github.com/reichlab/variant-nowcast-hub/actions/workflows/run-post-submission-jobs.yaml) are run (if the round_id in question is still in the 14 week window interim data window).

For normal hub operations, this is fine. However, if the target data jobs are run out of order (for example, a manual backfill run with an older nowcast_date), that would break the implicit assumption that `oracle.parquet` fill always reflect the most recent nowcast_date.

We decided that partitioning oracle outputs by sequence_date makes them too onerous to access. So we can't guarantee that overwriting the file won't happen, but we can add sequence_as_of information to the oracle output files as a breadcrumb.

### time series

The [RFC for Hubverse time series target data formats](https://github.com/reichlab/decisions/blob/main/decisions/2025-02-27-rfc-time-series-target-data.md) was approved on 2025-02-28 🎉 .

Based on the new time series guidance, we decided to:

- Partition time series target data on`as_of`, where `as_of` represents the `sequence_as_of` date
- Use a secondary partition of `nowcast_date`
- Explicitly include the above partition fields as columns in the time series parquet files
- Remove `tree_as_of` from the time series data (there is no immediate use case, and we're able to get the tree_as_of date from the `created_at` field in the`modeled_clades` json ([for example](https://github.com/reichlab/decisions/blob/main/decisions/2025-02-27-rfc-time-series-target-data.md)))

## Definition of done

- [ ] The `oracle.parquet` files generated by `get_target_data.py` will contain an `as_of` column that represents `sequence_as_of`
- [ ] The time series parquet files are partitioned by as_of (aka sequence_as_of) and then by nowcast_date
- [ ] The current time series parquet column `sequence_as_of` will be renamed to `as_of`
- [ ] The current time series parquet column `tree_as_of` will be removed
- [ ] The hub's [`target-data` README](https://github.com/reichlab/variant-nowcast-hub/blob/main/target-data/README.md) will contain a data dictionary for both time series and oracle output target data formats

bsweger:
Closes #296 

[This can be reviewed commit by commit]

This PR updates the organization and content of the variant-nowcast-hub's target data files as described on the above ticket. It also adds a data dictionary to the target-data README.

The following change wasn't part of the above ticket, but involves a data type change (thus, we want to apply it before re-creating all of the hub's target data and interim files):

- change the datatype of `oracle_value` and `observation` from float64 to int64, since those columns reflect counts.


rogersbw (approved):
Looks good

PR #411, opened 2025-03-21
Preds coming soon.

breesie:
Can I submit the preds for this week to see if they pass?

### DIFF
--- a/model-metadata/CADPH-CATaMaran.yml
+++ b/model-metadata/CADPH-CATaMaran.yml
@@ -2,7 +2,7 @@
 team_name: "California Department of Public Health"  # No more than 50 characters
 team_abbr: "CADPH"  # Up to 16 alphanumeric characters or underscores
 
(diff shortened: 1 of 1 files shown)


IsaacMacarthur:
I don't think you can submit your predictions for the week of 2025/03/19 because one of the checks is that the predictions are submitted before 8:00 pm on that Wednesday. However, if you want to test your predictions locally for the week of 2025/03/19, that is possible: you can use [the hubValidations package](https://github.com/Infectious-Disease-Modeling-Hubs/hubValidations), to check if the submission file will pass. The specific function you want will be hubValidations::validate_model_data.

You can also make a test submission to the hub for the week of 2025/03/26 starting Monday once the modeled clades list is out; if you do want to make a test submission like this, just add something like "do not merge" to the PR title so I know not to merge it.

If you run into any problems submitting, just let me know and I'm happy to help any way I can. It's really exciting that the model is almost ready!

I'll merge in the model-metadata file so that is ready for you.

IsaacMacarthur (approved):

PR #421, opened 2025-03-27
Standardize Python dependencies in requirements.txt

bsweger:
At the beginning, `variant-nowcast-hub` had a single Python script. So we specified Python dependencies using inline metadata, and it was good.

Then we added another Python script and also used inline metadata for dependencies, and it was pretty good.

Then we added a third Python script, this one with tests. And it used inline metadata for dependencies, but we also added a requirements.txt file to support the test runner. And then it was decent but not the best.

Then we added dependabot, which doesn't work on dependencies specified via metadata.

So this PR removes all script metadata in favor of using [`--with-requirements`](https://docs.astral.sh/uv/reference/cli/#uv-run--with-requirements) with every `uv run` command.

The `uv run` command is now a bit clunky, but these scripts are designed to run via CI and not by humans.

### DIFF
--- a/.github/workflows/create-modeling-round.yaml
+++ b/.github/workflows/create-modeling-round.yaml
@@ -47,7 +47,7 @@ jobs:
         env:
           GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          uv run get_clades_to_model.py
+          uv run --with-requirements requirements.txt get_clades_to_model.py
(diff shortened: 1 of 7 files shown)


zkamvar (approved):
This makes sense to me. It's a shame that you can't specify the requirements as metadata and have them be updated, but the plus side is that this change now allows the scripts to be run by something other than uv without manually setting up a requirements file (as I understand it).

PR #422, opened 2025-03-28
Adding  LANL predictions

MKupperman:
Adding LANL-CovTransformer predictions for  prediction date.\n> The views expressed here are those of the authors and does not necessarily represent the official views of the National Institutes of Health, Los Alamos National Laboratory, or the US Government.


MKupperman:
Guess this is finally working again.

IsaacMacarthur (approved):

PR #423, opened 2025-03-28
Adding Baseline model submissions

IsaacMacarthur:
Adding the submission files for previous rounds for the baseline line model as well as adding the model metadata for the baseline model. Ignore most of these many commits; it's just merges from main. The only files changed are related to the baseline model.


trobacker (commented):
LGTM! I'll let Nick do a final review.

PR #430, opened 2025-04-02
April 2 nowcast

Ehsan-suez:


IsaacMacarthur:
Hi UGA team, your submission for the week failed the hub checks. Looking at the file, it looks like you didn't model 23A, which is one of the required clades for the week. If you are having trouble modeling that specific clade, you are not alone. That clade also gave our model some trouble, and the solution that we used was to delete the one observation of 23A and add an observation at a time previous, say 100 or 150 days ago which fixed the problem.

IsaacMacarthur (approved):

Issue #433, opened 2025-04-04
Update clade inclusion criteria and capture additional metadata


bsweger:
**Background**

A combination of low sample volumes and our current clade "inclusion for modeling" criteria resulted in clade `23A` being included in the most recent modeling round. This, in spite of the fact that only one sequence in the prior 3 weeks was assigned to 23A, and prior to 2025-03-17, 23A hadn't appeared in a sample since 2024-08-18.

At today's Variant Nowcast Hub operational meeting, we decided to slightly alter the inclusion criteria as follows:

> At least 1% (in one week) and at least 2 sequences (across all weeks)

In other words, any clade included in the "clades to model" list because it met the 1% threshold will be removed if there are not at least two occurrences of it.

We also decided to capture some additional information in the weekly "clades to model metadata" ([example of current metadata](https://github.com/reichlab/variant-nowcast-hub/blob/main/auxiliary-data/modeled-clades/2025-04-02.json#L14-L27)):

- total number of sequences considered (i.e., the prior 3 weeks)
- total number of sequences for each clade included in the round

**Definition of done**

- [ ] Add a test suite to the current [`get_clades_to_model.py` script](https://github.com/reichlab/variant-nowcast-hub/blob/main/src/get_clades_to_model.py) (this script used to be part of cladetime and was tested there, but the tests didn't make it to the variant-nowcast-hub repo)
- [ ] Update inclusion criteria as stated above, including updated tests
- [ ] Update README (include date change went into effect)
- [ ] Add information to `modeled-clades` metadata as stated above

PR #437, opened 2025-04-08
Add test harness for the weekly get_clades_list process

bsweger:
Before we begin the logic changes outlined in #433, we should have formal and automated tests in place for the current
logic that generates the weekly list of clades to model.

This PR adds those tests and can be reviewed commit by commit:

- Small refactor for easier testing
- Add the tests
- Create a new workflow that run the tests whenever there's a change to `src/get_clades_to_model.py`


rogersbw (approved):
The changes make sense to me and solve the issue, code looks good and test all pass on my end!

PR #440, opened 2025-04-09
Update Clade Inclusion Criteria

trobacker:
This PR amends `get_clades_to_model.py` to use the new clade inclusion criteria outlined in #433. 

The added logic is to determine which clades have at least 2 counts (sequences) over the last several weeks (default threshold is 3 weeks). Then this list of clades is used to filter before applying the 1% threshold. 

I verified that this resolved the 23A issue we saw for nowcast date 2025-04-02.

### DIFF
--- a/src/get_clades_to_model.py
+++ b/src/get_clades_to_model.py
@@ -70,6 +70,7 @@ def get_clades(
 
     # Based on the most recent sequence collection date and the threshold_weeks parameter,
     # determine the minimum sequence collection date to consider when generating the clade list.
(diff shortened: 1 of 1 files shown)


trobacker:
A lingering thought on my initial commit: it boils down to counting if there's at least 2 counts per clade in the threshold weeks. So, it does catch the issue of 23A since it had only 1 count and occurrence. However, if there was one occurrence with 2 counts, it will pass through. Do we want this to be more stringent? e.g. at least 2 occurrences of each clade?

PR #449, opened 2025-04-15
Adding 2025-04-16 LANL predictions

MKupperman:
Adding LANL-CovTransformer predictions for 2025-04-16 prediction date.

> The views expressed here are those of the authors and does not necessarily represent the official views of the National Institutes of Health, Los Alamos National Laboratory, or the US Government.


MKupperman:
Investigating 24C.

MKupperman:
There we go. New clade import tool working as intended.

PR #453, opened 2025-04-16
CATaMaran update

Chipdelmal:
First time running the pipeline, so hopefully it is in the right shapes


IsaacMacarthur:
Hi CADPH, your submission for the weeks failed a few checks; looking at the error logs, it looks like the model output file was placed in the wrong folder, the CADPH-CATaLog folder not the CADPH-CaTaMaran folder. Also it looks like the wrong clades were modeled. It looks like 23A, and 24B, and 24C were modeled, which were not part of this week's modeled clade list. If you want to try and fix the submission file, you can resubmit even though the deadline has passed. Let me know if there is anything I can do can do to help.

IsaacMacarthur (approved):

Issue #455, opened 2025-04-17
unscored-location-dates file not created for nowcast date 2025-04-16


bsweger:
The GitHub actions runner aborted during the weekly run of the [run-post-submissions-job workflow](https://github.com/reichlab/variant-nowcast-hub/blob/main/.github/workflows/run-post-submission-jobs.yaml) (the other part of that job, creating new and interim target data files, completed successfully).

Logs from the failed run:  https://github.com/reichlab/variant-nowcast-hub/actions/runs/14505433627/job/40726979294

bsweger:
Even when re-running the job with verbose logging enabled, there's not a lot of information to go on:

```
##[debug]Re-evaluate condition on job cancellation for step: 'Create file of sequence counts by location and date 🦠'.
##[debug]Skip Re-evaluate condition on runner shutdown.
Error: The operation was canceled.
##[debug]System.OperationCanceledException: The operation was canceled.
##[debug]   at System.Threading.CancellationToken.ThrowOperationCanceledException()
##[debug]   at GitHub.Runner.Sdk.ProcessInvoker.ExecuteAsync(String workingDirectory, String fileName, String arguments, IDictionary`2 environment, Boolean requireExitCodeZero, Encoding outputEncoding, Boolean killProcessOnCancel, Channel`1 redirectStandardIn, Boolean inheritConsoleHandler, Boolean keepStandardInOpen, Boolean highPriorityProcess, CancellationToken cancellationToken)
##[debug]   at GitHub.Runner.Common.ProcessInvokerWrapper.ExecuteAsync(String workingDirectory, String fileName, String arguments, IDictionary`2 environment, Boolean requireExitCodeZero, Encoding outputEncoding, Boolean killProcessOnCancel, Channel`1 redirectStandardIn, Boolean inheritConsoleHandler, Boolean keepStandardInOpen, Boolean highPriorityProcess, CancellationToken cancellationToken)
##[debug]   at GitHub.Runner.Worker.Handlers.DefaultStepHost.ExecuteAsync(IExecutionContext context, String workingDirectory, String fileName, String arguments, IDictionary`2 environment, Boolean requireExitCodeZero, Encoding outputEncoding, Boolean killProcessOnCancel, Boolean inheritConsoleHandler, String standardInInput, CancellationToken cancellationToken)
##[debug]   at GitHub.Runner.Worker.Handlers.ScriptHandler.RunAsync(ActionRunStage stage)
[...]

PR #459, opened 2025-04-17
Address polars streaming shenanigans

bsweger:
This PR is the first attempt to address #455 

The Polars team is in the process creating a new streaming engine and has been issuing warnings that the old one (that the `get_location_date_counts.py` script was using) will become more stable over time.

It's not entirely clear if the recent GitHub workflow failure of this script was caused by the streaming engine, but this the hub's only Python script that uses it, so this PR removes that option to see what happens.

The PR also updates the README's Polars example code to use the new streaming syntax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -422,7 +422,7 @@ lf = pl.scan_parquet(
 )
 
 # use a collect operation to materialize the LazyFrame into a DataFrame
-model_output = lf.collect(streaming=True)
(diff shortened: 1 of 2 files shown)


trobacker (approved):
Looks good to me! 
I like the comment, "Streaming is a great Polars feature, but it's still in development. Thus, we should avoid it in the hub's operational workflow unless absolutely necessary."

PR #461, opened 2025-04-17
Add information about administrative workflows to src/README

### LINKED ISSUE #460: Update src/README.md to include a section about workflows
**Background**

The variant nowcast hub's `src/` directory has gotten a lot bigger in the past few months....so many scripts! The [README](https://github.com/reichlab/variant-nowcast-hub/blob/main/src/README.md) describes many of them, but it doesn't have information about which scripts are automated via GitHub workflows and how to re-run them if needed.

**Definition of done**

- [ ] For automated scripts, the README.md linked above contains information about which workflow runs them
- [ ] This information includes how/whether to re-run if there's a problem and what to consider

bsweger:
Closes #460 

Today's failure of the run-post-submissions-jobs workflow was a reminder that we don't have good documentation on whether or not it's safe to run the variant nowcast hub's automated jobs.

This PR adds some brief information about the behavior of administrative workflows and how they handle round_ids when they are re-run.

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -162,4 +162,43 @@ To run the script manually:
 
 ## Workflows
 
-Coming soon!
+Many of the scripts in `variant-nowcast-hub/src` are run via scheduled
+[GitHub workflows](https://docs.github.com/en/actions/about-github-actions/understanding-github-actions#workflows),
+usually when a new round is ready to open or when a round closes for submissions.
+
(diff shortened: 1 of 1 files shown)


IsaacMacarthur (approved):
This all sounds good to me! I've tried re-running failed workflows before and it's good to have clear language on what workflows can be re-run without causing problems.

trobacker (approved):
LGTM! Thanks for clarifying these aspects!

PR #493, opened 2025-05-07
Adding the hub-baseline submission for 05-07-2025

IsaacMacarthur:
Adding the hub-baseline submission for 05-07-2025; note this week's submission is based on 150 days of training data because of fitting weirdness.


trobacker (approved):
Adding an extra note about this week's submission: based on 150 days of training data instead of the usual 50 for the Hub-baseline.

PR #495, opened 2025-05-07
Uploading May7th CADPH-CAT nowcasts

Chipdelmal:


IsaacMacarthur:
The hub checks are not working at the second, but the submission passed the checks locally, so it is being merged in.

IsaacMacarthur (approved):

PR #542, opened 2025-06-04
Update Scoring Procedure Code

trobacker:
Update scoring procedure to exclude dates that had data available during the nowcast period as per the Hub's description.

To check this, one can run the example command then cross-references with the appropriate `auxillary-data/unscored-location-dates` and verify the counts > 0 days are excluded.


IsaacMacarthur (approved):
The changes seem to do want we want them to do! I did some tests and in all cases it excluded the days with previous counts, so it looks good to me!

PR #551, opened 2025-06-11
Updating Summary Plot Visualizations

trobacker:
Updating point/model color scheme and adding nowcast date visualization.

You can test this using commands within the repo directory such as:
`source("./src/plot_summary_graphs.R")` then 
`plot_summary_graphs(hub_path = "./", model_output_file = "UMass-HMLR/2025-06-11-UMass-HMLR.parquet", s3_data_date = "2025-06-09", save_path = "~/Desktop/", baseline_clade = "25A")`

### DIFF
--- a/src/plot_summary_graphs.R
+++ b/src/plot_summary_graphs.R
@@ -11,8 +11,8 @@
 
 #'
 #' @examples
-#' plot_summary_graphs(model_output_file = "LANL-CovTransformer/2024-11-13-LANL-CovTransformer.parquet",
-#'                     s3_data_date = "2024-11-11")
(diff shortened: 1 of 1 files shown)


IsaacMacarthur (changes requested):
The function as written was not working for me; the problem was the s3_data_date. Both plot_one_location, plot_one_clade, and save_plots  use a variable called s3_data_date, but they don't take it in their function definition, or define it inside the function, so it was throwing errors for me. I was able to fix this by adding  s3_data_date variable in the function definition for all three functions, and setting this to be the s3_data_date that the user passes in. I can commit my fix if you want, but there might be other (better) ways to fix this , if you want to try something else. Just let me know either way!

IsaacMacarthur (approved):
LGTM!

PR #573, opened 2025-06-26
Register Bedford Lab hierarchical MLR model

huddlej:
Adds metadata for the Bedford Lab's hierarchical MLR model (https://nextstrain.org/sars-cov-2/forecasts).


trobacker:
Welcome!! There's a small validation error with the `.yml` file. The model name should be no more than 50 characters. As is, it is 57 with spaces. A quick modification should do the trick..!

trobacker (approved):
Success!

PR #585, opened 2025-07-07
Add Brier Score to Scoring Functions

trobacker:
Add Brier score (scaled), change column name for `es_score`, and clean up some code. You can look at example scores by running the example code in `model_scoring_functions.R`.


IsaacMacarthur (approved):
LGTM, the function now works with LANL and UGA submissions!

Issue #775, opened 2025-10-18
Run post-submission jobs failing for dates before 2025-09-28


IsaacMacarthur:
The Run post-submission jobs action is failing for dates before 2025-09-28, see https://github.com/reichlab/variant-nowcast-hub/actions/runs/18561495649 and https://github.com/reichlab/variant-nowcast-hub/actions/runs/18616395224/job/53082327083 for examples.

The error seems to be based on Nextstrain removing old versions of the files/ncov/open/metadata_version.json file; based on my look into the S3 bucket, it seems as though the oldest available public version of the file is for 2025-09-28.

You can see this by running the following R code, for example:
library(aws.s3)
s3HTTP(
  verb = "GET",
  bucket = "nextstrain-data",
  query = list(versions = "", prefix = "files/ncov/open/metadata_version.json"),
  anonymous = TRUE,
  region = "us-east-1"
)

PR #814, opened 2025-11-07
Weekly update of scores.tsv

trobacker:
Weekly update of evaluation scores.


IsaacMacarthur:
I have a question about this PR; It looks like the scores being added are for the nowcast date 2025/08/06, but we don't have finalized oracle output for that date because of #775. Are you just using the partially observed data?

trobacker:
Thanks, @IsaacMacarthur. Indeed, the last time we had updated target data was `2025-10-07`, as shown in the history of the oracle output files. The target data gets updated each week, though up to 90 days past, so running scoring didn't have trouble - it's just not using the fully validated data. 

So, I suppose we shouldn't even score nowcast submissions past `as.Date("2025-10-07")-90` = `2025-07-09`? I'll modify the scores to include only up to that nowcast submission `2025-07-09`, as that's the last 90-day window we can trust. Then we can run scores we've missed out on once we resolve the metadata issue. 

Thanks for questioning this - I was too automatic in running it!

trobacker:
@IsaacMacarthur `scores.tsv` now includes only `nowcast_date`'s up to and including `2025-07-09`.

PR #886, opened 2025-12-16
Issue 883: add coverage

kaitejohnson:
*Update*
Reran to include the "interval range" column which we will want to compute interval coverage.  Ready for review. 

This PR does the following:
- creates a script which calculates the coverage by target date, location, and clade for each  model and nowcast date, using the same logic used for the energy scores to produce 100 expected observed counts for each posterior draw (10,000 samples total for each nowcast date-target date-clade-model
- it by default computes the 50th and 90th intervals (so 0.05, 0.25, 0.5, 0.75. 0.95) prediction intervals. I chose not to add more to this because this makes the file that much longer -- however, if we want more prediction intervals I can rerun this 
- from those expected counts, compute the coverage for a specified set of quantiles
- for those without samples, add an error set of rows
- bind all the parquet files with coverage together across the nowcast dates and models and clades

Minor difference:
- I saved the file as a parquet not a TSV as the tsv was too big -- this one is bigger because we have coverage by clade and also because there are multiple quantiles per forecast unit. 

In general, I tried to follow the format and style of the scores PRs so it will be easier to perform the step of calculating these regularly as part of the GitHub actions evaluation workflow. For now, I haven't proposed any regular GitHub actions workflows-- this just creates the coverage.parquet retrospectively across all the nowcast dates and models with evaluation data available.

### DIFF
--- a/auxiliary-data/scores/README.md
+++ b/auxiliary-data/scores/README.md
@@ -63,8 +63,8 @@ This file will be updated weekly as target data is available to compute coverage
 - `location`: the two-letter abbreviation for a US state, including DC and PR for Washington DC and Puerto Rico.
 - `clade`: the alpha-numeric clade designation
 - `quantile_level`: the probability level of the quantile (e.g., 0.25, 0.75). Each prediction interval is defined by two quantiles. 
-- `interval_range`: the nomincal coverage level of the prediction interval (e.g., 50 for a 50% interval formed by the 0.25 and 0.75 quantiles).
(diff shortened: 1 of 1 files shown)


IsaacMacarthur (approved):
This all looks good to me! One suggestion that might be nice would be to add a README that explains what all of the columns are in the coverage parquet file; I was looking through and I was not entirely sure what they all were. Not a deal breaker though, if we want the coverages merged in quickly

IsaacMacarthur (dismissed):
LGTM! Thank you for adding the README!

trobacker (approved):
I think this all looks great! Thanks for adding the coverage functionality to the existing infrastructure. 

I left a comment on the README, but it's more just me processing the quantile coverage deviation metric rather than suggesting any changes. 
Approved! ✅

nickreich:
Ok - I get that with the csv size. makes sense. I defer to you and @trobacker and @IsaacMacarthur about placement of the files. I don't feel that strongly about this, and since it's well-documented, maybe it's just fine as is.

PR #925, opened 2026-01-02
Adding 2025-12-31 LANL predictions

MKupperman:
Adding LANL-CovTransformer predictions for 2025-12-31 prediction date.

> The views expressed here are those of the authors and do not necessarily represent the official views of the National Institutes of Health, Los Alamos National Laboratory, or the US Government.


MKupperman:
@IsaacMacarthur seems CI was not updated to reflect the 12/19 email for this submission week. Job runner is trying to apply the 1/7 due date check incorrectly.

IsaacMacarthur:
Sorry about that! I'll just bypass the checks and merge it in

IsaacMacarthur (approved):

IsaacMacarthur:
Because of some failing checks that did not have to do with the submission being late, which were not noticed, this file was improperly merged in. In particular, the file has incorrect nowcast and target dates. Because the file has the wrong nowcast and target dates, the file would cause issues for evaluation, and thus we are removing it to prevent any more downstream issues, see #947

PR #1136, opened 2026-04-28
Adding the baseline model for the week

IsaacMacarthur:
Adding the baseline model for the week of 2026-04-29


IsaacMacarthur:
The baseline model this week had convergence issues if we used 50 days of training data, so this model was fit using 70 days of training data.

trobacker (approved):

PR #1150, opened 2026-05-05
Adding the baseline model for the week

IsaacMacarthur:
Adding the baseline model for the week of 2026-05-06


IsaacMacarthur:
For this week only we had to use 80 days of training data to get a stable fit

trobacker (approved):

PR #1199, opened 2026-05-31
Fix GITHUB_PAT format error caused by new ghs_ token format

IsaacMacarthur:
GitHub changed the GITHUB_TOKEN format on 2026-04-27 from a short ~40-char ghs_ string to a longer ~520-char JWT. The old format accidentally passed gh R package PAT validation (ghs_ matched the gh[pousr]_ character class and fit within length bounds). The new format fails on both length and character constraints.

Group A (validate-submission, validate-config, cache-hubval-deps): Move GITHUB_PAT from job-level to setup-r-dependencies step only, so pak retains authenticated access while the R script execution steps use gh's GITHUB_TOKEN fallback.

Group B (create-ensemble, create-modeling-round, update-scores-weekly): Remove GITHUB_PAT from R script execution steps only. The setup-renv step retains GITHUB_PAT since renv does not call validate_gh_pat() and needs auth for package downloads.

Tested locally against PR #1193 (UGA-multicast 2026-05-27): broken state reproduces the format warning; fix passes all 23 validation checks cleanly.
Co-authored by Claude

### DIFF
--- a/.github/workflows/cache-hubval-deps.yaml
+++ b/.github/workflows/cache-hubval-deps.yaml
@@ -14,8 +14,6 @@ jobs:
   build-deps-cache-on-main:
     if: ${{ github.repository_owner == 'reichlab' }}
     runs-on: ubuntu-latest
-    env:
-      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
 
     steps:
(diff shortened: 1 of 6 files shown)


trobacker (approved):
🤞🏻

